PR #2, opened 2016-08-17
Add usage docs

Lewuathe:
I added initial usage documentation.

- [x] mix-in binding
- [x] local variable binding
- [x] creating design
- [x] binding singleton and tag


xerial:
Nice!

PR #5, opened 2016-08-22
[WIP] Object life cycle management

xerial:


xerial:
@Lewuathe Although this LifeCycleManagement has some room for improvement, let me merge it first since this code covers all of the functionalities of wvlet-inject.

PR #6, opened 2016-08-23
Session and LifeCycleManager refactoring

xerial:
- Split Session and LifeCycleManager
- Hide private method from Session (e.g., Session.get, etc.)
- Fix AirframeMacro to build objects (e.g. non-static, abstract types) correctly
- Make LifeCycleManager customizable by chaining (or wrapping) LifeCycleEventHandler


xerial:
@Lewuathe  Now lifecycle has three stages: init (called upon object injection) -> start -> shutdown

PR #7, opened 2016-08-24
Add Scala check style

Lewuathe:
- [x] Add scala check style plugin in compile phase.
- [x] Fix warnings issued with current check style config.


xerial:
Do you have IntelliJ code format configuration to ensure this?

PR #9, opened 2016-08-24
Deprecate Design.build[X]

xerial:
Directly building objects from Design is not preferrable, since there would be no reliable way to find a session used for generating these objects. 

Deprecate this method, and force users to use `Design.newSession.build[X]`.


xerial:
cc: @Lewuathe

PR #10, opened 2016-08-25
Fix singleton holder

xerial:


src/main/scala/wvlet/airframe/AirframeMacros.scala
```diff
@@ -32,37 +32,55 @@ object AirframeMacros extends LogSupport {
...
-      //
-      // = Non static type
-      // If X is non static type (= local class or trait),
-      // we need to instantiate it first in order to populate its $outer variables
-      if(!a.isStatic || (a.isStatic && !a.isAbstract && hasPublicDefaultConstructor)) {
+      if(shouldInjectSession) {
         q"""{
-             ${c.prefix}.getOrElseUpdate[$t]((new $t {
-               protected[this] def __current_session = ${c.prefix}
+          val session = ${c.prefix}
```
xerial: We need to explicitly bind ${c.prefix} to a variable, because it can be complex chain of method calls.

src/main/scala/wvlet/airframe/Airframe.scala
```diff
@@ -16,11 +16,5 @@ import wvlet.log.LogSupport
 
 object Airframe extends LogSupport {
   def newDesign: Design = Design.blanc
-
-  private var _currentSession : Option[Session] = None
```
xerial: Deprecated current session holder, since user always need to call design.newSession

src/main/scala/wvlet/airframe/SessionImpl.scala
```diff
@@ -57,9 +57,10 @@ private[airframe] class SessionImpl(sessionName:Option[String], binding: Seq[Bin
     val t = ObjectType.ofTypeTag(ev)
     binding.find(_.from == t) match {
       case Some(SingletonBinding(from, to, eager)) =>
-        singletonHolder.getOrElseUpdate(to, {
-          registerInjectee(to, obj)
-        }).asInstanceOf[A]
+        singletonHolder.getOrElseUpdate(from, registerInjectee(to, obj)).asInstanceOf[A]
+      case Some(InstanceBinding(form, obj)) =>
+        // Instance is already registered in init
+        obj.asInstanceOf[A]
```
xerial: Missing this line was the cause why @PostConstruct is called multiple times for singletons

xerial:
cc: @Lewuathe

PR #11, opened 2016-08-25
Improve injection performance with compile-time code generation

xerial:
This PR translates bind[X]:
```
val x = bind[X] 
```
into the following code **if X is non abstract type** and has the default constructor:
```
val x = new X { protected[this] __current_session = (session) }
```

This is much faster than using runtime reflection to instantiate trait.


xerial:
cc: @Lewuathe

Issue #15, opened 2016-08-25
Remove binding from Design


xerial:
For convenience of modifying existing Design

xerial:
Fixed by #16

PR #16, opened 2016-08-26
#15 Add Design.remove[A]

xerial:


src/main/scala/wvlet/airframe/Design.scala
```diff
@@ -23,14 +23,14 @@ import scala.reflect.runtime.{universe => ru}
 /**
   * Immutable airframe design
   */
-class Design(val binding: Seq[Binding]) extends LogSupport {
+case class Design(binding: Vector[Binding]) extends LogSupport {
```
xerial: Using Vector for better performance of appending

xerial:
cc: @Lewuathe

PR #19, opened 2016-08-26
Fix scalastyle

xerial:
There is no reason for us to restrict the line width to 100 (though this is used in Databricks). line width 100 tends to introduce unnecessary line wraps and breaks the code flow. 

And also we need to allow using 4 and 5 ass non magic number for writing provider codes.


xerial:
cc: @Lewuathe

Issue #25, opened 2016-09-08
Avoid runtime code compilation for performance improvement


xerial:
When binding a trait:
```scala
    .bind[StatementHandler].toEagerSingletonOf[DefaultStatementHandler]
```

If the trait has no Session, we need to dynamically embed a Session object to it:
```
2016-09-08 10:10:41-0700 trace [wvlet.airframe.SessionImpl] Search bindings of StatementHandler  - (SessionImpl.scala:105)
2016-09-08 10:10:41-0700 trace [wvlet.airframe.SessionImpl] Found a singleton for StatementHandler: DefaultStatementHandler  - (SessionImpl.scala:115)
2016-09-08 10:10:41-0700 trace [wvlet.airframe.SessionImpl] Search bindings of DefaultStatementHandler  - (SessionImpl.scala:105)
2016-09-08 10:10:41-0700 trace [wvlet.airframe.SessionImpl] No binding is found for DefaultStatementHandler  - (SessionImpl.scala:141)
2016-09-08 10:10:41-0700 trace [wvlet.airframe.SessionImpl] buildInstance DefaultStatementHandler, stack:List(DefaultStatementHandler, StatementHandler)  - (SessionImpl.scala:148)
2016-09-08 10:10:41-0700 trace [wvlet.airframe.SessionImpl]
Compiling a code for embedding Session to DefaultStatementHandler:
new (wvlet.airframe.Session => Any) {
  def apply(session:wvlet.airframe.Session) = {
    new com.treasuredata.prestobase.proxy.StatementHandler.DefaultStatementHandler {
      protected def __current_session = session
    }
  }
}    - (SessionImpl.scala:184)
2016-09-08 10:10:42-0700 trace [wvlet.airframe.Session] Checking a session for class __wrapper$1$7ff11152d6594d33b9c958a6a4b57872.__wrapper$1$7ff11152d6594d33b9c958a6a4b57872$$anon$2$$anon$1  - (Session.scala:113)
```

This uses reflection to compile the code at runtime, and slows down the initialization process.
To avoid this, we should generate a code with Scala macros in advance when binding a singleton in the above example.

xerial:
Added a code to check runtime compilation time and it took around 0.5 ~ 1.0 sec: 
{code}
Compiling a code for embedding Session to JSONWebLogGenerator:
new (wvlet.airframe.Session => Any) {
  def apply(session:wvlet.airframe.Session) = {
[...]

PR #27, opened 2016-09-13
Pre-compile session injection code

xerial:
For #25.

- [x] Avoid reflection when instantiating a trait (an abstract class, but has no abstract methods, which will be just a Java interface in Scala 2.12) 
- [x] Add sub-projects for using macros inside Airframe code (e.g., Design will need to use macro to generate a code template, but macro cannot be called within the same project) 
- [x] Design.bind[X].toSingleton[Y]. A default session injection code (factory) for Y needs to be generated at compile-time instead of using reflection.


xerial:
Done. This fix improved the dependency injection speed significantly (from 0.5 ~ 1 sec. to almost no time)

PR #29, opened 2016-09-19
Improve CircleCI testing

xerial:
- Improved dependency fetch speed with https://github.com/alexarchambault/coursier


xerial:
@Lewuathe This improves dependency download speed significantly. Try it in other projects :)

PR #42, opened 2017-01-12
Add comparison of DI frameworks

xerial:


xerial:
cc: @Lewuathe @takezoe

PR #44, opened 2017-03-07
[WIP] surface: Reflection-free object schema Implementation

xerial:
- [x] Add Surface.of[X] as a replacement of ObjectSchema
- [x] Support Scala.js, which does not have runtime reflection
- [x] Support aliased types for binding objects instead of using tagged type (@@)
- [x] Primitive types
- [x] Array types
- [x] Scala Collection types
  - [x] Seq
  - [x] List
  - [x] Map
  - [x] Set
  - [x] ParSeq
  - [x] Tuple
  - [x] Option
  - [x] Either
  - [x] Try
- [x] Aliased types
- [x] JavaCollection types
- [x] Standard Types
  - [x] java.io.File
  - [x] java.time objects
- [x] Extract method information (MethodSurface)
- [x] Extract method annotation
- [x] Extract class annotation
- [ ] Instantiate objects from Surface


xerial:
To implement this, https://github.com/adamw/macwire/blob/master/macros/src/main/scala/com/softwaremill/macwire/internals/ConstructorCrimper.scala this code would be useful.

xerial:
Moved the development to https://github.com/wvlet/surface. I'll close this ticket

PR #45, opened 2017-03-13
Migrate to Surface

xerial:
Migrating to https://github.com/wvlet/surface, reflection-free object surface library

- [x] Remove dependency to object-schema
- [x] Fix test case compilation errors
   - [ ] ~Fix Surface's annotation extractor~
- [x] Deprecate @PreDestroy, @PostConstruct initialization 
- [ ] ~Add companion apply() method finder to Surface~
- [x] Fix Serialization issue of Surface
- [x] Implement Session finder without using reflection
- [ ] ~Share test code with Scala.js~


xerial:
Major changes include:
- ObjectType -> Surface
- Deprecate embedding Session as a parameter or providing from a function. We are comfortable even if we cannot embed Session as a parameter. 
- Deprecate @PostConstruct and @PreDestroy life cycle hook, because if we use these annotations in a trait, these lifecycle methods can be overwritten unintentionally. 
- It would work with Scala.js, I'll work on that in another ticket.

xerial:
cc: @Lewuathe

Lewuathe:
👍

PR #49, opened 2017-04-06
Cross building for Scala JVM and Scala.js

xerial:
- [x] Cross build for JVM and js
- [x] Fix Scala.js test cases


xerial:
@Lewuathe FYI:
Splitting build process into test:fastOptJS and test was necessary to avoid errors like https://github.com/typelevel/cats/issues/1242

PR #60, opened 2017-05-23
[WIP] Object builder

xerial:
https://github.com/wvlet/config requires building objects from Properties. Surface should support that.

- [x] Add zero value generator for populating default values
- [ ] Add ObjectBuilder with .set(param name, value) method
- [ ] Support nested object parameter set .set(path, value)


xerial:
Done in #62

PR #61, opened 2017-05-23
Stop using PhantomJS which causes JSCom has been closed error

xerial:


xerial:
This doesn't resolve the issue at all.

PR #62, opened 2017-05-26
Add reflection-based Surface generator to support Surface(ru.Type)

xerial:
This is necessary for using https://github.com/wvlet/config with the latest version of Airframe without changing the code so much.


xerial:
This PR adds reflection-based surface generation and ObjectBuilder, which is necessary in wvlet-config to override config object with property values.

PR #66, opened 2017-06-03
Release 0.15

xerial:


xerial:
This will be addressed in #67  and #68

Issue #63, opened 2017-06-03
Extensively use runtime-reflection


xerial:
It seems runtime-reflection (TypeTag) is flexible enough to provide necessary functionality for Airframe and Surface. An idea here is switching implementation of Surface for JVM and Scala.js (or Scala native).

- Switching Surface implementation between JVM and Scala.js
- Add Airframe interface that can provide runtime type information

xerial:
Done in #67

PR #67, opened 2017-06-04
Surface factory

xerial:
For #63. Reorganizing project structure
- Surface JVM: Mostly use runtime-reflection to build Surface. Favors fast compilation
- Surface JS: Using Scala Macros to generate Surface because reflection is unavailable in Scala.js
- Rename Surface.of[X] -> wvlet.surface.of[X] 
  - To provide different implementations for Scala JVM / JS
  - Initially I tried to provide different Surface object implementation, but Scala forbids defining Surface trait and Surface object in different files.


xerial:
@Lewuathe Unlike the initial motivation of Surface, the reflection-free implementation doesn't have much benefit for us. And also I can find a better implementation to extract type parameters without reading ScalaSig (which caused deadlock in td-spark). So with this PR, Surface mostly uses runtime-type parameter (ru.Type) which is provided by Scala compiler.

PR #71, opened 2017-06-05
Merge wvlet-config as airframe-config

xerial:
- wvlet-config is often used with Surface and Airframe. Merging wvlet-config to airframe is good for better maintainability. 
- wvlet-config version is already reaching 1.0, while airframe is now 0.15. To avoid version conflict, renamed it to airframe-config.


xerial:
@Lewuathe I'll merge this since any change to surface may cause test failure in config project, so we should maintain them in the same repository.

PR #73, opened 2017-06-06
 Use RuntimeSurface to convert Map type element

Lewuathe:
Objects which can be build from Buffer is constructed with one element type like Array type. It is necessary to convert each element to Tuple2 type in case of Map type.

### DIFF
--- a/airframe-config/src/main/scala/wvlet/config/YamlReader.scala
+++ b/airframe-config/src/main/scala/wvlet/config/YamlReader.scala
@@ -18,9 +18,9 @@ import java.{util => ju}
 import org.yaml.snakeyaml.Yaml
 import wvlet.log.LogSupport
 import wvlet.log.io.IOUtil._
-import wvlet.surface.reflect.ObjectBuilder
 import wvlet.surface
 import wvlet.surface.Surface
+import wvlet.surface.reflect.ObjectBuilder
 
 import scala.collection.JavaConverters._
 import scala.collection.immutable.ListMap
@@ -76,6 +76,10 @@ object YamlReader extends LogSupport {
           case al: java.util.ArrayList[_] =>
             for (a <- al.asScala) {
               builder.set(k.toString, a)
+            }
+          case map: java.util.Map[_, _] =>
+            for ((mk, mv) <- map.asScala) {
+              builder.set(k.toString, mk -> mv)
             }
           case _ =>
             builder.set(k.toString, v)
--- a/airframe-config/src/main/scala/wvlet/config/YamlReader.scala
+++ b/airframe-config/src/main/scala/wvlet/config/YamlReader.scala
@@ -72,18 +72,19 @@ object YamlReader extends LogSupport {
     val builder = ObjectBuilder(surface)
     if (prop != null) {
       for ((k, v) <- prop.asScala) {
-        v match {
-          case al: java.util.ArrayList[_] =>
-            for (a <- al.asScala) {
-              builder.set(k.toString, a)
-            }
-          case map: java.util.Map[_, _] =>
-            for ((mk, mv) <- map.asScala) {
-              builder.set(k.toString, mk -> mv)
-            }
-          case _ =>
-            builder.set(k.toString, v)
-        }
+        builder.set(k.toString, v)
+//        v match {
+//          case al: java.util.ArrayList[_] =>
+//            for (a <- al.asScala) {
+//              builder.set(k.toString, a)
+//            }
+//          case map: java.util.Map[_, _] =>
+//            for ((mk, mv) <- map.asScala) {
+//              builder.set(k.toString, mk -> mv)
+//            }
+//          case _ =>
+//            builder.set(k.toString, v)
+//        }
       }
     }
     builder.build.asInstanceOf[A]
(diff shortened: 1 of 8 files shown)


airframe-config/src/main/scala/wvlet/config/YamlReader.scala
```diff
@@ -69,20 +68,18 @@ object YamlReader extends LogSupport {
...
-      }
+    val factory = surface.objectFactory
+    if (factory.isDefined) {
+      val args = prop.asScala.map {
+        case (_, v: java.util.ArrayList[_]) => v.asScala
+        case (_, v: java.util.Map[_, _]) => v.asScala.toMap
+        case (_, v) => v
+      }.toSeq
+
+      factory.get.newInstance(args).asInstanceOf[A]
```
xerial: Note factory.newInstance(args) is order-sensitive.

xerial: Usually this is just calling the primary constructor of the class

airframe-config/src/main/scala/wvlet/config/YamlReader.scala
```diff
@@ -69,20 +68,18 @@ object YamlReader extends LogSupport {
 
   def bind[A](surface: Surface, prop: java.util.Map[AnyRef, AnyRef]): A = {
     trace(s"bind ${surface}, prop:${prop.asScala}")
-    val builder = ObjectBuilder(surface)
-    if (prop != null) {
-      for ((k, v) <- prop.asScala) {
-        v match {
-          case al: java.util.ArrayList[_] =>
-            for (a <- al.asScala) {
-              builder.set(k.toString, a)
```
xerial: This code is mapping object parameter key -> value

airframe-config/src/main/scala/wvlet/config/YamlReader.scala
```diff
@@ -69,20 +68,18 @@ object YamlReader extends LogSupport {
...
-            }
-          case _ =>
-            builder.set(k.toString, v)
-        }
-      }
+    val factory = surface.objectFactory
+    if (factory.isDefined) {
+      val args = prop.asScala.map {
```
xerial: This code looks ignoring keys specified in the map.

airframe-config/src/test/resources/classes.yml
```diff
@@ -0,0 +1,17 @@
+default:
+  classes:
+    - class1
+    - class2
+    - class3
+  class_assignments:
```
xerial: 👍

xerial:
I understand the reason why we need this, but I think extending ObjectBuilder to support setting Map type values is much more straight forward.

xerial:
I think if we add java.util.Map type mapping around here, we can use Map value: 
https://github.com/wvlet/airframe/blob/1f799b425dd27e1021c103b49ecef7b9e556e4d8/surface/jvm/src/main/scala/wvlet/surface/reflect/TypeConverter.scala#L76

Lewuathe:
> I think extending ObjectBuilder to support setting Map type values is much more straightforward.
Thanks. I'll try.

xerial:
It seems ObjectBuilder API is confusing for you, but there is a reason why set(path, value) needs to accept both collection element or collection itself. I'll add modify this PR

airframe-config/src/main/scala/wvlet/config/YamlReader.scala
```diff
@@ -77,6 +77,10 @@ object YamlReader extends LogSupport {
             for (a <- al.asScala) {
               builder.set(k.toString, a)
             }
+          case map: java.util.Map[_, _] =>
```
xerial: @Lewuathe This logic is a bit ad-hoc solution. What I meant in the previous comment a month ago is removing these mappings (for java arrays, maps) to ObjectBuilder side.

xerial: I'll work on this fix.

airframe-config/src/main/scala/wvlet/config/YamlReader.scala
```diff
@@ -72,14 +72,7 @@ object YamlReader extends LogSupport {
...
       for ((k, v) <- prop.asScala) {
-        v match {
-          case al: java.util.ArrayList[_] =>
-            for (a <- al.asScala) {
-              builder.set(k.toString, a)
-            }
-          case _ =>
-            builder.set(k.toString, v)
-        }
+        builder.set(k.toString, v)
```
xerial: Simplified the usage of object builder by supporting adding single element or collection.

surface/jvm/src/main/scala/wvlet/surface/reflect/ObjectBuilder.scala
```diff
@@ -110,50 +113,85 @@ trait StandardBuilder extends GenericBuilder with LogSupport {
...
-            case Some(Value(v)) =>
-              // remove the default value
-              holder.remove(name)
-            case _ => // do nothing
+      else {
+        trace(s"set path $path : $value")
+        if (path.isLeaf) {
+          val targetType = p.get.surface
+          trace(s"update value holder name:$name, valueType:$targetType (isArray:${isArray(targetType)}) with value:$value (${value.getClass})")
+          if (canBuildFromBuffer(targetType)) {
```
xerial: A reason we need to prepare ArrayHolder here is we will use ObjectBuilder to parse command line arguments. For example, when parsing 'command -a A1 -a A2 ...', we need to call set("a", "A1"), set("a", "A1") separetely when the result type is Config(a:Seq[String]).

xerial:
@Lewuathe I'll merge this and deploy a new version.

Lewuathe:
Thanks!

PR #98, opened 2017-08-24
Merge wvlet-log as airframe-log

xerial:


xerial:
@Lewuathe Migrated all commits from wvlet-log to make airframe a mono-repo of utilities. 
We need to remove "wvlet-log" dependencies after airframe 0.20.

PR #115, opened 2017-11-14
Upgrade to logback-core 1.2.3 and sbt-1.0.3

xerial:
For making it compatible with Presto 0.188, which uses logback-core 1.2.3

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -246,7 +246,7 @@ lazy val log =
     )
     .jvmSettings(
       libraryDependencies ++= Seq(
-        "ch.qos.logback" % "logback-core" % "1.1.7"
+        "ch.qos.logback" % "logback-core" % "1.2.3"
       )
     )
     .jsSettings(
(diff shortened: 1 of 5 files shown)


project/build.properties
```diff
@@ -12,4 +12,4 @@
 # limitations under the License.
 #
 
-sbt.version=1.0.2
+sbt.version=1.0.3
```
xerial: sbt-1.0.2 has a bug in finding resources and log file discovery generated in test case failes in 1.0.2.

build.sbt
```diff
@@ -246,7 +246,7 @@ lazy val log =
     )
     .jvmSettings(
       libraryDependencies ++= Seq(
-        "ch.qos.logback" % "logback-core" % "1.1.7"
+        "ch.qos.logback" % "logback-core" % "1.2.3"
```
Lewuathe: 👍

PR #116, opened 2017-11-14
Upgrade to Scala 2.12.4

xerial:


xerial:
cc: @Lewuathe Scala 2.12.4 has improved the compilation speed.

PR #124, opened 2017-12-02
Extract MessagePack Codec as airframe-codec

xerial:
- Minimum dependencies (e.g., except surface and log) to be a reusable interface. => future work: define msgpack-java based implementation elsewhere. #125 
- It would be better to remove the dependency to msgpack-java (for supporting Scala.js and Scala.native in future) => future work. Introduce Packer/Unpacker interface. #125 
- project name. airframe-msgpack, ariframe-msgcodec, or just airframe-codec? => airframe-codec


xerial:
#125 will be a follow-up.

xerial:
cc: @Lewuathe @j14159

PR #127, opened 2017-12-03
[WIP] airframe-msgpack: Pure-Scala MessagePack SPI + Implementation

xerial:
Design considerations: 
- pure-scala implementation for the use in Scala (JVM), Scala.js, and Scala.native
- Support offset based packer/unpacker (for better error handling without moving buffer internal cursors)
- Support marking on memory buffers (for holding memory region that contains a value)


airframe-msgpack/shared/src/main/scala/wvlet/airframe/msgpack/spi/Packer.scala
```diff
@@ -0,0 +1,53 @@
...
+import java.time.Instant
+
+/**
+  *
+  */
+/**
+  * Message Packer interface
+  */
+trait Packer {
+  def packNil: Packer
```
xuwei-k: `def packNil: this.type` is better?

xerial: good point. I'll use this.type

PR #140, opened 2018-01-25
Fix airframe-opts doc typo Launguer to Launcher.

kimutansk:
When trying example code, I found class name typo.


xerial:
Merged. Thanks for the fix!

Issue #151, opened 2018-02-23
Macros are not working as advertised for Scala.js


sumnulu:
Steps:
1- Turn off the logging with `Logger.rootLogger.setLogLevel(LogLevel.OFF)`
2- Compile with fullOptJS
3- XXXX-opt.js contains all the trace/warn/info/etc. methods with the full path of the source code inlined.

---

For some reason I was expecting the macro would wrap the log methods in **if** blocks with **final** values inside which compiler can remove the whole block. 

Are there any way to remove logging code completely? 
Especially inlined fullpath of the source files can be a problem for the web.

---
```javascript
    var level = $m_Lwvlet_log_LogLevel$INFO$();
    if (this$3.$$undlog__p1__ju_logging_Logger().isLoggable__ju_logging_Level__Z(level.jlLevel$1)) {
        this.logger__Lwvlet_log_Logger().log__Lwvlet_log_LogLevel__Lwvlet_log_LogSource__O__V(
$m_Lwvlet_log_LogLevel$INFO$(), new $c_Lwvlet_log_LogSource().
init___T__T__I__I("/Users/sumnulu/Projects/log1/js/src/main/scala/web/Main.scala", 
"Main.scala", 31, 7), "Main run test 123")
    }
```

xerial:
@sumnulu airframe-log uses macros for wrapping logging code within the if-block, so it will not remove the logging code. The point is that this logging code will not be executed at run-time, so basically it removes the runtime overhead. 

Scala macros generates these codes at **compile-time**, so for totally removing logging code (e.g., for production usage), we may need to use some environment variable to change the behavior of airframe-log macros. It means we cannot recover the removed logging code after compilation. 

- A) Allowing compile-time log level configuration (e.g., suppress generation of logging code below INFO level)
- B) Embedding only the leaf file name of the source file

I think these two are technically possible. B) is easy. A is not so important for Scala, but I understand it's good for Scala.js.

xerial:
Fixed in #152

PR #152, opened 2018-02-25
[log] Add option to disable code gen

sumnulu:
ref: https://github.com/wvlet/airframe/issues/151

### DIFF
--- a/airframe-log/README.md
+++ b/airframe-log/README.md
@@ -262,6 +262,13 @@ if(logger.isDebugEnabled) {
 Log message String generation will not happen unless the debug log is effective. 
 Scala macro is also used for finding source code location (LogSource).
 
+### To disable Logging at compile-time
+It is possible to disable the generation of logging statements at compile time.
+Add `scalacOptions += "-Xmacro-settings:wvlet.log.disable.[REPLACE_WITH_LEVEL]"` compiler option to `build.sbt` 
+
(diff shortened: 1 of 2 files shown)


xerial:
@sumnulu Thanks for the PR. Considering the characteristic of this compile-time only option, having less code coverage for this patch is OK.

airframe-log/README.md
```diff
@@ -262,6 +262,13 @@ if(logger.isDebugEnabled) {
 Log message String generation will not happen unless the debug log is effective. 
 Scala macro is also used for finding source code location (LogSource).
 
+### To disable Logging at compile-time
+It is possible to disable the generation of logging statements at compile time.
+Add `scalacOptions += "-Xmacro-settings:wvlet.log.disable.[REPLACE_WITH_LEVEL]"` compiler option to `build.sbt`
```
xerial: The meaning of REPLACE_WITH_LEVEL is not clear without looking LocMacros code. Can we use a more clearer naming? For example, SUPPRESS_BELOW = {ERROR, WARN, INFO, DEBUG}: 

If SUPPRESS_BELOW = INFO, then DEBUG, TRACE log code generation will be suppressed.

sumnulu: Thanks for the feed back. Is it OK now?

xerial: Yes. LGTM now

PR #168, opened 2018-04-29
Add JDK9/10 to the build matrix

xerial:


xerial:
Test passed after clearing some caches for Scala.js project.

PR #206, opened 2018-08-12
[WIP] Add Simple HTTP REST framework

xerial:
- [x] Using jax-rs annotations for defining REST endpoints
- [x] Matching http requests to target functions
- [ ] Populate function arguments 
- [x] Add finagle binding


xerial:
Merging since the basic framework is done. Function call mapping will be implemented in another PR.

PR #207, opened 2018-08-15
[WIP] airframe-http: rest mapping

xerial:
- [x] http request -> msgpack -> function call


airframe-http/src/main/scala/wvlet/airframe/http/EndPoint.java
```diff
@@ -0,0 +1,35 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.ElementType.TYPE;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+@Retention(RUNTIME)
+@Target({METHOD, TYPE})
+public @interface EndPoint
```
seratch: (Just from my curiosity) 
Which is more common, `Endpoint` or `End Point`? `Endpoint` is more familiar to me.

xerial: Thanks. We should use Endpoint :)

PR #218, opened 2018-08-22
Update airframe-metrics document and test for 0 duration

takezoe:
Time window representation examples in airframe-metrics's documentation are wrong about 0 duration. I fixed documentation and added test to confirm the difference of 0[hdM] and 1[hdM].


xerial:
Thanks for the update!

PR #220, opened 2018-08-23
Support quarter (q) duration

xerial:
- There is no ChronoUnit.Quarter in JDK, so created ~TimeUnit~ TimeWindowUnit classes to use Quarter unit.
- The beginning of the quarter is January. But note that the definition of a quarter can be different depending on company culture.


xerial:
Thanks for checking!

PR #227, opened 2018-08-28
use canonical name for arguments of method surface on MethodCallBuilder

yajirobee:
`MethodCallBuilder` fail to use default value if argument name includes
capital letter. Therefore, airframe-opts's sub-command don't use default value
for such argument as follows.

```
scala> class SampleCommand() {
     |   @command()
     |   def sample(@option(prefix = "-a") aB: Int = 1) = {
     |     println(aB)
     |   }
     | }
defined class SampleCommand

scala> val l = Launcher.of[SampleCommand]
l: wvlet.airframe.opts.Launcher = wvlet.airframe.opts.Launcher@544a71e6

scala> l.execute(Array("sample"))
0
```

It is because `ObjectBuilder` tries to find arguments with canonical name.
https://github.com/yajirobee/airframe/blob/305e71afb5bda93a961d75d93a035c90e8b13905/airframe-surface/jvm/src/main/scala/wvlet/surface/reflect/ObjectBuilder.scala#L201


xerial:
Good catch! Thanks for the fix.

PR #246, opened 2018-09-08
Fix MyApp constructor

j5ik2o:
I think the MyApp constructor is wrong. Am I wrong? Please confirm...


xerial:
You are right. The doc is wrong. Thanks for the fix!

PR #263, opened 2018-09-26
Add bindFactory[I=>X] usage

xerial:


xerial:
cc: @Lewuathe @takezoe

Issue #269, opened 2018-09-29
Multi-cloud/region config override support


xerial:
airframe-config should be able to support multi-cloud environment.

xerial:
We will not work on this since config override is too complicated

PR #276, opened 2018-10-08
[WIP] Multicloud config

xerial:
- [x] tag-based config
- [ ] Allow grouping based on custom tags

Related #269


xerial:
The semantics of such config override is quite complex. We should find a better way to scope configuration.

PR #293, opened 2018-11-01
 Add untruncate notation to offset of time window

takezoe:
I propose a notation to untruncate offset in time window like:

- `0M/2018-09-02)` -> `[2018-09-01 00:00:00-0700,2018-09-02 00:00:00-0700)`
- `-1M/2018-09-02 12:34:56)` -> `[2018-08-01 00:00:00-0700,2018-09-02 12:34:56-0700)`

In short, if offset is an exact date and ends with `)`, it's not truncated.

My remaining concern about this proposal is how it should work for the future time windown like `+1M/2018-09-02)`. My current implementation recognizes it as `[2018-09-02 00:00:00-0700,2018-10-01 00:00:00-0700)` because the tail `)` simply means untrancating, but this behavior may be confusing.


docs/src/main/tut/docs/airframe-metrics.md
```diff
@@ -117,3 +117,5 @@ Here are several examples of relative time window strings when the current time
 |`-1h/2017-01-23 01:00:00`| last hour from the given offset (hour) | `2017-01-23 00:00:00-0700` | `2017-01-23 01:00:00-0700`|
 |`-1h/2017-01-23 01:23:45`| last hour from the given offset (hour) | `2017-01-23 00:00:00-0700` | `2017-01-23 01:00:00-0700`|
 |`-1M/2017-01-23 01:23:45`| last month from the given offset (hour) | `2016-12-01 00:00:00-0700` | `2017-01-01 00:00:00-0700`|
+|`0M/2017-01-23)`| from beginning of the month to a give offset | `2017-01-01 00:00:00-0700` | `2017-01-23 00:00:00-0700`|
```
xerial: to a _given_ offset

docs/src/main/tut/docs/airframe-metrics.md
```diff
@@ -117,3 +117,5 @@ Here are several examples of relative time window strings when the current time
...
+|`+1M/2017-01-23 01:23:45)`| from a give offset to end of the month | `2017-01-23 01:23:45-0700` | `2017-02-01 00:00:00-0700`|
```
xerial: a _given_

xerial:
- Introducing the notion of `untruncate` seems good because the behavior of forward (+1M) and backward(-1M) time vectors becomes consistent.

xerial:
Merged into master. Thanks!

docs/src/main/tut/docs/airframe-metrics.md
takezoe: Oops. Thanks for fixing! 🙇

docs/src/main/tut/docs/airframe-metrics.md
takezoe: 🙇

PR #354, opened 2019-01-11
Fix in-trait injection example

takezoe:
In-train injection example at the web site doesn't work because `name` isn't given to `Database`.


xerial:
Thanks for the fix.

PR #358, opened 2019-01-18
Allows to use bindFactory outside of wvlet.airframe package

### LINKED ISSUE #357: bindFactory causes compilation error
```scala
package com.github.takezoe
import wvlet.airframe._

trait Logger {
  val tag: String = bind[String]
}

trait LoggerService {
  val factory = bindFactory[String => Logger]
}
```

This code causes a following compilation error.

```
value createNewInstanceOf in trait Session cannot be accessed in wvlet.airframe.Session
```

If above code is defined under `wvlet.airframe` package, it can be compiled.

takezoe:
Closes #357


xerial:
@takezoe Could you add some test here to see this actually resolve the issue?
https://github.com/wvlet/airframe/tree/master/airframe/src/test/scala/example

PR #359, opened 2019-01-19
Enhance loan pattern for AutoCloseable

takezoe:
I propose following two changes for the loan pattern for `AutoCloseable`:

- Ignore resource closing errors
- Add two resources version of `withResource`

I think that the two resources version is useful for common usecases like copying `InputStream` to `OutputStream`.


airframe-control/src/main/scala/wvlet/airframe/control/Control.scala
```diff
@@ -13,15 +13,38 @@
...
+    }
+  }
+
+  def withResources[R1 <: AutoCloseable, R2 <: AutoCloseable, U](resource1: R1, resource2: R2)(
+      body: (R1, R2) => U): U = {
+    try {
+      body(resource1, resource2)
+    } finally {
+      closeQuietly(resource1)
+      closeQuietly(resource2)
```
xerial: I think we should close `resource2` even if `resource1.close` throw an faital error. How about nesting one more try-catch?

takezoe: like d8b157182281714d4345c811ecae92b40e5ff776?

xerial:
OK. LGTM. Go ahead to merge

PR #360, opened 2019-01-19
Add name conversion functionallity to CName

takezoe:
- Refactored existing code
  - Not use `return`
  - Avoid cast in `CName#equals()`
  - Use `HashMap` instead of `WeakHashMap` in Scala.js
- Added name conversions from natural name
  - `snakeCase`
  - `dashCase`
  - `upperCamelCase`
  - `lowerCamelCase`


xerial:
@takezoe I think we should use a regular HashMap for Scala.js. For example by adding Compat.newWeakHashMap = ...

takezoe:
@xerial Got it. Thanks for your advise!

xerial:
ok. LGTM. It’s OK to merge this by yourself as a committer :)

PR #363, opened 2019-01-22
Change airframe-control to a Scala JVM project

xerial:
This is necessary to add IO related code to airframe-control. 

This is OK since Scala.js will not use Retry.


xerial:
cc: @takezoe

PR #368, opened 2019-01-24
Bump to Fluency 2.0.0

takezoe:
This pull request upgrades Fluency which is used in airframe-fluentd to 2.0.0.

Since configuration object has been abolished in Fluency 2.0.0, I created a case class for configuration in airframe-fluentd. But I'm not sure if this is the best solution because when new configuration items are added to Fluency in the future, we need to catch up them one by one.


xerial:
Please run `sbt scalafmtSbt` to pass the code style check at Travis CI.

airframe-fluentd/src/main/scala/wvlet/airframe/fluentd/FluencyClient.scala
```diff
@@ -14,20 +14,50 @@
...
 import org.komamitsu.fluency.{EventTime, Fluency}
 import wvlet.log.LogSupport
 import wvlet.airframe._
 
 case class FluencyConfig(
     // Use the extended EventTime timestamps
     // https://github.com/fluent/fluentd/wiki/Forward-Protocol-Specification-v1#eventtime-ext-format
-    useExtendedEventTime: Boolean = false,
-    fluencyConfig: Fluency.Config = new Fluency.Config()
+    useExtendedEventTime: java.lang.Boolean = false,
```
xerial: I think we should use Scala primitive values. In `newFluency` function, we should use Boolean.box(x) to bridge with Java.

airframe-fluentd/src/main/scala/wvlet/airframe/fluentd/FluencyClient.scala
```diff
@@ -14,20 +14,50 @@
...
+    maxBufferSize: java.lang.Long = null,
```
xerial: Option[Long] = None. 

Then 
```
builder.setMaxBufferSize(maxBufferSize.getOrElse(null)) 
or
maxBufferSize.foreach { x => builder.setMaxBufferSize(x) }
```

xerial:
I love type safe configuration. So having config case class is good in general.

For future changes,
- Adding new configuration is less harmful, because airframe-fluentd users just cannot use the new configuration.
- Having deprecated configuration is more problematic since it will cause runtime errors because some method will be missing in a newer version of fluency. For example even if airframe-fluentd is using Fluency 2.0 api, the classpath dependency can have Fluency 2.1. In this case MethodNotFound error will be thrown.  So wrapping this exception and show the warning message upon error will add more safety.


In terms of API design, I think we should create FluencyBuilder:
```
Fluency.newBuilder
.withBufferChunkInitialSize(...)
.withFileBackupDir(...)
.withErrorHandler(...)
.withSSL // or .noSSL
.build
```

takezoe:
Thanks for your advise.

Warning missing methods is good idea. I think Fluency configuration should be supplid via DI in the airframe-fluentd design. So I would like to create a builder for configuration object rather than a builder for FluencyClient.

takezoe:
Modified configuration interface to make possible to give `FluencyConfig` in natural Scala way.

airframe-fluentd/src/main/scala/wvlet/airframe/fluentd/FluencyClient.scala
```diff
@@ -22,12 +24,63 @@ case class FluencyConfig(
...
+    setConfiguration(builder, "errorHandler", _.setErrorHandler(fluencyConfig.errorHandler))
+    setConfiguration(builder, "senderMaxRetryCount", _.setSenderMaxRetryCount(fluencyConfig.senderMaxRetryCount))
+    setConfiguration(builder, "ackResponseMode", _.setAckResponseMode(fluencyConfig.ackResponseMode))
+    setConfiguration(builder, "sslEnabled", _.setSslEnabled(fluencyConfig.sslEnabled))
+
+    builder.build(fluentdConfig.host, fluentdConfig.port)
+  }
+
+  // Wrap a function to warn deprecated configuration items
+  private def setConfiguration(builder: FluencyBuilderForFluentd,
```
xerial: 👍

airframe-fluentd/src/main/scala/wvlet/airframe/fluentd/FluencyClient.scala
```diff
@@ -22,12 +24,63 @@ case class FluencyConfig(
...
+    maxBufferSize: Long = 512 * 1024 * 1024,
+    bufferChunkInitialSize: Int = 1024 * 1024,
+    bufferChunkRetentionSize: Int = 1024 * 1024 * 4,
+    bufferChunkRetentionTimeMillis: Int = 1000,
+    flushIntervalMillis: Int = 600,
+    fileBackupDir: String = null,
+    waitUntilBufferFlushed: Int = 60,
+    waitUntilFlusherTerminated: Int = 60,
+    jvmHeapBufferMode: Boolean = false,
+    errorHandler: ErrorHandler = null,
```
xerial: I think we have no way to set this errorHandler value via Yaml-based configuration (e.g., using airframe-config) unless providing default implementations (e.g., defautErrorHandler, etc.) 

We have two approaches: passing Fluency configuration or passing Fluency itself in DI:
```
val fluency: Fluency = bind{ errorHandler: FluencyErrorHandler => Fluency.newClient.withErrorHandler(errorHandler).build } 
```

As I commented before https://github.com/wvlet/airframe/pull/368#issuecomment-457337067, using type safe builder interface is most flexible for such configuration.

If we need to use config object, we still can combine both approaches: 
```
bind {config:FluencyConfig => Fluency.newBuilder.withConfig(config).withErrorHandler(...).build }
```

I think the configuration parameters we will use frequently include:
- useExtendedEventTime
- maxBufferSize
- flushIntervalMillis
- jvmHeapBufferMode = true (I'm not sure how safe it is to use DirectByteBuffer in fluency) 
- ackResponseMode = true/false (This depends on whether we can accept missing data)
- sslEnabled = true (For data security)

We also need to determine these two parameters programmatically:
- fileBackupDir (e.g., `{spark.tempdir}/{app id}/fluency`)
- errorHandler 

So instead of having nulls in config, how about adding a simple Fluency builder which accepts the above basic configuration set + withXXX methods for backup dir + error handler?

takezoe: I see. In any case, it's difficult choice in design. What about providing pre-configured `Design` as follows?
```scala
package object fluentd {
  def withFluency(host: String = "127.0.0.1",
                  port: Int = 24224,
                  useExtendedEventTime: Boolean = false,
                  maxBufferSize: Long = 512 * 1024 * 1024,
                  flushIntervalMillis: Int = 600,
                  jvmHeapBufferMode: Boolean = true,
                  ackResponseMode: Boolean = true,
                  sslEnabled: Boolean = true): Design = ...
}
```
Users can simply override above properties as they needed. But still we need to consider how to configure `fileBackupDir` and `errorHandler` programmatically. Builder might be necessary anyway.

airframe-fluentd/src/main/scala/wvlet/airframe/fluentd/FluencyClient.scala
```diff
@@ -14,28 +14,39 @@
...
+  private val fluencyClientConfig = bind[FluencyClientConfig]
+  private val fluency: Fluency = bind { (fluencyClientConfig: FluencyClientConfig) =>
+    val builder = new FluencyBuilderForFluentd()
+    builder.setMaxBufferSize(fluencyClientConfig.maxBufferSize)
+    builder.setFlushIntervalMillis(fluencyClientConfig.flushIntervalMillis)
+    builder.setJvmHeapBufferMode(fluencyClientConfig.jvmHeapBufferMode)
+    builder.setAckResponseMode(fluencyClientConfig.ackResponseMode)
+    builder.setSslEnabled(fluencyClientConfig.sslEnabled)
+    builder.setFileBackupDir(fluencyClientConfig.fileBackupDir)
+    builder.setErrorHandler(fluencyClientConfig.errorHandler)
```
xerial: Does it accept null values?

xerial:
ok. Let's merge it and try using it. We will figure out whether this API is good enough or not.

airframe-fluentd/src/main/scala/wvlet/airframe/fluentd/FluencyClient.scala
takezoe: Yes, builder's default value is null. If it's null, builder doesn't do anything.

takezoe:
#404 is successor of this pull request.

PR #371, opened 2019-01-27
Not ignore resource closing exception in withResource

### LINKED ISSUE #370: Control.withResource should throw an exception upon close?
withResource implementation consumes the exception. It seemed convenient, but because of this I always need to worry about the error handling when using this method. For example, if closing FileOutputStream fails, it means the written data was lost. In this case, we should not consume the exception. 

@takezoe 
I'd like to change this method to throw an exception:
```
  def withResource[R <: AutoCloseable, U](resource: R)(body: R => U): U = {
    try {
      body(resource)
    } finally {
      closeQuietly(resource)
    }
  }
```

takezoe:
Closes #370

### DIFF
--- a/airframe-control/src/main/scala/wvlet/airframe/control/Control.scala
+++ b/airframe-control/src/main/scala/wvlet/airframe/control/Control.scala
@@ -13,8 +13,6 @@
  */
 package wvlet.airframe.control
 
-import scala.util.control.NonFatal
-
 /**
   *
   */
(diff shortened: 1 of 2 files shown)


xerial:
Thanks for the fix. We need to fix the code format with sbt scalafmt

takezoe:
Oops, I often forget to do that...

xerial:
Thx

PR #374, opened 2019-02-01
Get ConfigHolders directly from Design

Lewuathe:
Every time ConfigHolder is needed, it's necessary to initialize the session from the design for now. 
ConfigHolder can be obtained directly from the design without additional overhead.

### DIFF
--- a/airframe-config/src/main/scala/wvlet/airframe/config/package.scala
+++ b/airframe-config/src/main/scala/wvlet/airframe/config/package.scala
@@ -42,7 +42,7 @@ package object config {
     }
 
     def bootstrapWithConfigProcessor(configProcessor: Config => Unit): Design = {
-      configProcessor(getConfigOrEmpty)
+      configProcessor(currentConfig)
       d
     }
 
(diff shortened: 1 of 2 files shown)


airframe-config/src/main/scala/wvlet/airframe/config/package.scala
```diff
@@ -42,16 +42,24 @@ package object config {
...
         case dc: DesignOptionsWithConfig =>
-          dc.config
+          Some(dc.config)
         case _ =>
-          Config()
+          None
+      }
+    }
+
+    def getConfigOrEmpty: Config = {
```
xerial: `currentConfig` might be better to distinguish with Option[Config]

xerial:
OK. Just change the method name.

airframe-config/src/main/scala/wvlet/airframe/config/package.scala
Lewuathe: 👍

xerial:
@Lewuathe btw, I recommend using your personal fork of airframe to send PR, because Travis CI will run twice for the airframe branch and PR

Lewuathe:
Oh, I should have. Okay.

Issue #373, opened 2019-01-31
Outer variable is not visible within `design.build[X] { x => ... }` block


xerial:
```scala
    val helloDesign = "hello"
    val d = newDesign
      .bind[String].toInstance(helloDesign)

    d.build[String] { x =>
      helloDesign // compile error
    }
```
This code works in Scala 2.11, but it doesn't in Scala 2.12.

The macro definition of build[X]:
https://github.com/wvlet/airframe/blob/129868ad8376011ef259c3935b4833c3ba9a0834/airframe-macros/src/main/scala/wvlet/airframe/AirframeMacros.scala#L410-L420

xerial:
Fixed in #376

PR #376, opened 2019-02-04
#373: design.build[X] { ... } closure scope

xerial:
It seems closure handling has changed after Scala 2.12 and local variables cannot be found inside the passed closure. #373

 In Scala 2.11 there is no problem.


xerial:
I still don't know why a local variable cannot be seen in the passed closure. It works in Scala 2.11, but started to fail in Scala 2.12.

Issue #377, opened 2019-02-05
airframe-vcr


xerial:
Add a module for recording HTTP request/responses

xerial:
Implemented in #380

PR #380, opened 2019-02-06
airframe-http-recorder

xerial:
#377 

- [x] Rename to airframe-http-recorder
- [x] Fix a bug in unescaping JSON strings
- [x] Add test


airframe-http-recorder/src/main/scala/wvlet/airframe/http/recorder/HttpRecordStore.scala
```diff
@@ -0,0 +1,117 @@
...
+
+  def find(request: Request): Option[HttpRecord] = {
+    val rh = requestHash(request)
+
+    // If there are multiple records for the same request, use the counter to find
+    // n-th request, where n is the access count to the same path
+    val counter  = requestCounter.getOrElseUpdate(rh, new AtomicInteger())
+    val hitCount = counter.getAndIncrement()
+    connectionPool.queryWith(
+      // Get the
```
takezoe: Imcompleted comment

airframe-http-recorder/src/main/scala/wvlet/airframe/http/recorder/HttpRecordStore.scala
```diff
@@ -0,0 +1,117 @@
...
+
+  override def close(): Unit = {
+    connectionPool.stop
+  }
+
+  /**
+    * Compute a hash key of the given HTTP request.
+    * This value will be used for DB indexes
+    */
+  def requestHash(request: Request): Int = {
```
takezoe: This method can be private?

airframe-http-recorder/src/main/scala/wvlet/airframe/http/recorder/HttpRecorder.scala
```diff
@@ -0,0 +1,115 @@
...
+                              recordTableName: String = "record",
+                              // Specify the port to use. The default is finding an available port
+                              private val port: Int = -1,
+                              // A filter for customizing HTTP request headers to use for generating database keys.
+                              // For example, we should remove headers that depends on the current time, etc.
+                              headerExcludes: String => Boolean = HttpRecorder.defaultHeaderExclude,
+                              fallBackHandler: Service[Request, Response] = HttpRecorder.defaultFallBackHandler) {
+
+  def sqliteFilePath  = s"${folder}/${sessionName}.sqlite"
+  lazy val serverPort = if (port == -1) IOUtil.unusedPort else port
```
takezoe: Just curious, why `sqliteFilePath` is a method despite `serverPort` and `destHostAndPort` are `lazy val`?

airframe-http-recorder/src/main/scala/wvlet/airframe/http/recorder/HttpRecordStore.scala
```diff
@@ -0,0 +1,117 @@
...
+    connectionPool.executeUpdate(s"delete from ${recordTableName} where session = '${recorderConfig.sessionName}'")
+  }
+
+  private val requestCounter = scala.collection.mutable.Map.empty[Int, AtomicInteger]
+
+  def resetCounter: Unit = {
+    requestCounter.clear()
+  }
+
+  def find(request: Request): Option[HttpRecord] = {
```
takezoe: `findNext` might be better in this case to show this method has side effect.

takezoe:
@xerial I left some tiny comments (or questions) on this pull request. Please take a look if you have a chance.

airframe-http-recorder/src/main/scala/wvlet/airframe/http/recorder/HttpRecorder.scala
xerial: Both can be val. IOUtil.unusedPort will return an available port at the moment, so its initialization should be delayed as much as possible.

airframe-http-recorder/src/main/scala/wvlet/airframe/http/recorder/HttpRecordStore.scala
xerial: sure. I'll make it a private

airframe-http-recorder/src/main/scala/wvlet/airframe/http/recorder/HttpRecordStore.scala
xerial: 👍 good catch

airframe-http-recorder/src/main/scala/wvlet/airframe/http/recorder/HttpRecordStore.scala
xerial: Agreed. I'll change the name

airframe-http-recorder/src/main/scala/wvlet/airframe/http/recorder/HttpRecorder.scala
xerial: A reason I chose def would be sqliteFilePath can be used only once, so it saves the memory space a little bit by making it a def. 

The use of `lazy val` is usually for avoiding the same computation.

takezoe: I got it! Thanks for your answer!

Issue #417, opened 2019-03-04
airframe-http: Routing issue


takezoe:
I defined two endpoints as:
	
```scala
@Endpoint(path = "/v1")
trait ResourceManagerService {
  @Endpoint(path = "/:cluster/workers/task", method = HttpMethod.GET)
  def workersTask(cluster: String) = {
    ...
  }

  @Endpoint(path = "/:cluster/workers/status", method = HttpMethod.GET)
  def workersStatus(cluster: String) = {
    ...
  }
}
```

Then run the following command for testing:

```
$ curl http://localhost/v1/hoge/workers/task
```

But `workersStatus()` was called against my expectation.

takezoe:
Now I'm investigating the cause of this behavior.

takezoe:
The current implementation of `Router` seems to allow to define the parameter (`:name`) only at the last component of the path.

https://github.com/wvlet/airframe/blob/71cea45762c82dff4bad8ec01540dd2cfb1c2126/airframe-http/src/main/scala/wvlet/airframe/http/Router.scala#L30-L37

https://github.com/wvlet/airframe/blob/71cea45762c82dff4bad8ec01540dd2cfb1c2126/airframe-http/src/main/scala/wvlet/airframe/http/Router.scala#L86-L91

PR #418, opened 2019-03-05
airframe-http: Allow more flexible http path routing

takezoe:
The current implementation of airframe-http's router allows to define the parameter only at the last component of the path. This pull request allows more flexible http path routing like:

- `/:param/hoge`
- `/:param1/hoge/:param2`

See also #417

### DIFF
--- a/airframe-http/src/main/scala/wvlet/airframe/http/Router.scala
+++ b/airframe-http/src/main/scala/wvlet/airframe/http/Router.scala
@@ -31,9 +31,16 @@ class Router(val routes: Seq[Route]) {
     routes
       .find { r =>
         r.method == request.method &&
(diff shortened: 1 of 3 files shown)


xerial:
Thanks for the fix!

takezoe:
Thanks for merging.

Remaining concern is that it's unclear which endpoint is called if multiple endpoints are matched with the request path. We might have to improve the path matching logic to have the clear and stable  behavior for conflicted path definitions.

For example, it might be better to give priority to the concrete path definitions (e.g. `/user/info`) than the parameterized path definitions (`/:id/info`). Currently, it depends on the order of definition of methods.

xerial:
I think we should run some conflicting path check when creating Route instances.

Issue #420, opened 2019-03-06
bind[X].toInstance(X(...)) shows compilation error


xerial:
This usually doesn’t happen, so we need to reproduce this error. 

cc: @takezoe

takezoe:
Yes, I will try to create a test case to reproduce.

takezoe:
I created a minimum testcase to reproduce the compilation error in #422.

PR #422, opened 2019-03-07
Minimum testcase to reproduce compilation error in toInstance binding

takezoe:
This pull request is for showing a minimum testcase to reproduce compilation error in toInstance binding which was reported in #420.


takezoe:
Doesn't CI work for draft pull requests?

xerial:
@takezoe I will send a commit to fix this issue.

PR #423, opened 2019-03-07
Fix bind instance error

xerial:
Resolves #422 and #420


xerial:
@takezoe The cause of compilation error was, the locally created case classes were not visible inside the generated code. I've fixed the scope of macro-generated code.

takezoe:
@xerial Thanks for the fix!

PR #427, opened 2019-03-10
airframe-sql

xerial:
@takezoe I would like to maintain sql parser as a part of airframe. This PR is an import from the commits of msgframe.

Several reasons I want to include airframe-sql:
- SQL will be useful to filter data streams generated by airframe-codec
- SQL data processing for msgpack data streams needs to be tightly coupled with msgpack parser implementation (airframe-msgpack) and memory management (airframe-canvas). 
- By combining TimeWindow (e.g., -1d, -1w in airframe-metrics) with SQL, we can write a query generator for time-series data.


takezoe:
@xerial Makes sense. In addition, I had felt it would become easier to maintain if airframe-sql is in the same repository because it allows us to focus on airframe repository.

Issue #425, opened 2019-03-09
[Proposal] airframe-control: Implicit classes offering universal syntax sugar for Parallel


takezoe:
The current code which uses `Parallel` is:

```scala
import wvlet.airframe.control.Parallel

// for Seq
val result = Parallel.run(source, parallelism = 3) { i =>
  ...
}

// for Iterator
val result = Parallel.iterate(source, parallelism = 3) { i =>
  ...
}
```

My idea is adding implicit classes allowing above code to rewrite as follows:

```scala
import wvlet.airframe.control.Parallel.implicits._

// for Seq and Iterator
val result = source.parallelMap(parallelism = 3) { i =>
  ...
}
```

How do you think about this idea?

xerial:
I prefer using the base package object `import wvlet.airframe.control.parallel._` if this is for user convenience. 

For now adding this syntax sugar with implicit class is OK, because we don't need to chain parallel methods. If we need to support more concatenated operations that is sharing the same thread pool, we may need to introduce free-monad like operators (e.g., Parallel(seq).map(x => ...) ) to embed the execution context like thread pools.  

Additional ideas:
- In future, I'd like to add similar options with GNU parallel (e.g., non-order preserving output, pipe (merging stdio, etc.), task queue size). 
- Seq[X] can be Iterate or TraversableOnce, so switching Parallel.run or Parallel.iterate will not work as you might expect. I think we should add mapParallel and mapParallelIterate. This reminds me `Stream[X]` abstraction, which will be deprecated in Scala 2.13 because of its confusing interface with Seq.

takezoe:
Basic syntax sugar has been implemented in #429 as:

```scala
Seq(1, 2, 3).parallel.map(x => ...)
Seq(1, 2, 3).parallel.withParallelism(4).map(x => ...)
```

I'll keep working on other improvements for Parallel. I would like to make possible to share same thread pool (`ExecutorService` or `ExecutionContext`?) among concatenated map processes at least.

takezoe:
I tried to implement some patterns sharing `ExecutionContext` across multiple map processes but I was not able to find a better solution than Scala's standard `Future`. In fact, it became very similar to `Future` because `Future` already has filled our requirements.

xerial:
Now that #429 is merged, let me close this.

PR #429, opened 2019-03-11
airframe-control: Implicit classes for parallel map

takezoe:
See #425

### DIFF
--- a/airframe-control/src/main/scala/wvlet/airframe/control/parallel/package.scala
+++ b/airframe-control/src/main/scala/wvlet/airframe/control/parallel/package.scala
@@ -17,17 +17,38 @@ import scala.reflect.ClassTag
 
 package object parallel {
 
-  implicit class ParallelSeq[T](source: Seq[T]) {
-    def parallelMap[R: ClassTag](parallelism: Int = Runtime.getRuntime.availableProcessors())(f: T => R): Seq[R] = {
-      Parallel.run(source, parallelism)(f)
-    }
+  implicit class ToParallelSeq[T](source: Seq[T]) {
+    def parallel: ParallelSeq[T] = ParallelSeq(source)
   }
 
-  implicit class ParallelIterator[T](source: Iterator[T]) {
-    def parallelMap[R: ClassTag](parallelism: Int = Runtime.getRuntime.availableProcessors())(
-        f: T => R): Iterator[R] = {
-      Parallel.iterate(source, parallelism)(f)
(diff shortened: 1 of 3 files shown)


xerial:
@takezoe Code formatting is failed

airframe-control/src/main/scala/wvlet/airframe/control/parallel/package.scala
```diff
@@ -0,0 +1,32 @@
...
+ * limitations under the License.
+ */
+package wvlet.airframe.control
+
+import scala.reflect.ClassTag
+
+package object parallel {
+
+  implicit class ParallelSeq[T](source: Seq[T]) {
+    def parallelMap[R: ClassTag](parallelism: Int)(f: T => R): Seq[R] = {
```
xerial: Should we set the default parameter for this (e.g., the number of cores)?

airframe-control/src/test/scala/wvlet/airframe/control/ParallelTest.scala
```diff
@@ -97,6 +97,29 @@ class ParallelTest extends AirframeSpec {
       assert(List(Success(2), Failure(exception), Success(6)).forall(x => result.contains(x)))
     }
 
+    "be run for Seq using syntax sugar" in {
+      import wvlet.airframe.control.parallel._
+
+      val source = Seq(1, 2, 3)
+      val result: Seq[Int] = source.parallelMap(parallelism = 2){ x =>
```
xerial: I think this syntax is a little bit unintuitive because the first argument is parallelism configuration, not the function to process the input elements (e.g., parallelMap(f: A=>B))

xerial:
@takezoe 

How about this syntax?
```scala
source.withParallelism(2).map(...)
source..withParallelism(2).withThreadPool(...).map(...)
```

airframe-control/src/test/scala/wvlet/airframe/control/ParallelTest.scala
takezoe: OK, Let's consider more intuitive syntax following your comment below.

takezoe:
I updated implicit classes to support above syntax.

To support sharing a thread pool between chained map processes, we need additional changes in `Parallel` not only adding implicit classes. So I would like to work on that in the another pull request.

xerial:
Ok. `Seq(1, 2, 3).parallel.map(x => ...)` looks good.

PR #442, opened 2019-03-17
airframe-http-finagle: Downgrade to Finagle 19.2.0

### LINKED ISSUE #383: Automatically switch recording/replay mode

xerial:


xerial:
Merging this first with only Finagle downgrade.

PR #451, opened 2019-03-21
Use Scala 2.13.x + JDK11

xerial:
It seems the community build is failing because of Serialization bug of Scala 2.13 collection library. 
 - Vector is not serializable

This PR is for reproducing that bug, and finding a workaround.


xerial:
Actually it seems the serialization bug occurs in the nightly builds of Scala 2.13.x. 

I think we should try the latest Scala 2.13.x in another branch.

PR #457, opened 2019-03-24
[DO NOT MERGE] Upgrade to sbt-1.3.0-M2

xerial:


xerial:
sbt-1.3.0-M2 has a regression. We should wait a next version that fixes the class loader issue

xerial:
sbt 1.3.0-RC1 is released #497

PR #480, opened 2019-03-30
Fix JMX methods extraction

takezoe:
Currently, only first method of methods annotated with `@JMX` is registered as JMX attribute.


xerial:
Nice catch!

Issue #473, opened 2019-03-29
airframe-control: insufficient concurrency of parallel execution


xerial:
Need to investigate the cause. cc: @takezoe

takezoe:
As a result of investigation, it was not a problem of airframe-control.

By the way, I added trace logs and JMX monitoring in #481. We can use them for investigation at the next time.

PR #481, opened 2019-03-30
Add trace logs and JMX monitoring to Parallel

takezoe:
Even though #473 has been reported, I haven't find any problems in `Parallel` implementation yet. To find problem easily I would like to add trace log and JMX monitoring to `Parallel`.

### DIFF
--- a/airframe-control/src/main/scala/wvlet/airframe/control/Parallel.scala
+++ b/airframe-control/src/main/scala/wvlet/airframe/control/Parallel.scala
@@ -14,18 +14,34 @@
 package wvlet.airframe.control
 
 import java.util.UUID
-import java.util.concurrent.atomic.AtomicReference
+import java.util.concurrent.atomic.{AtomicLong, AtomicReference}
 import java.util.concurrent._
 
+import wvlet.airframe.jmx.{JMX, JMXAgent}
 import wvlet.log.LogSupport
 
-import scala.concurrent.duration.Duration
 import scala.reflect.ClassTag
 
 /**
   * Utilities for parallel execution.
   */
 object Parallel extends LogSupport {
+
+  class ParallelExecutionStats(
+      @JMX
+      val executionId: String,
+      @JMX
+      val parallelism: Int,
+      @JMX
+      val total: String,
+      requestQueue: LinkedBlockingQueue[_],
+      counter: AtomicLong
+  ) {
+    @JMX
+    def runningWorkers: Int = parallelism - requestQueue.size()
+    @JMX
+    def count: Long = counter.longValue()
+  }
 
   private class ResultIterator[R](queue: LinkedBlockingQueue[Option[R]]) extends Iterator[R] {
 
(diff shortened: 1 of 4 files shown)


airframe-control/src/main/scala/wvlet/airframe/control/Parallel.scala
```diff
@@ -46,17 +65,27 @@ object Parallel {
     * @param f Function which processes each element of the source collection
     * @return Collection of the results
     */
-  def run[T, R: ClassTag](source: Seq[T], parallelism: Int = Runtime.getRuntime.availableProcessors())(
+  def run[T, R: ClassTag](source: Seq[T], parallelism: Int = Runtime.getRuntime.availableProcessors(), jmxAgent: Option[JMXAgent] = None)(
```
xerial: Instead of passing jmxAgent, I think we should use the default agent.

airframe-control/src/main/scala/wvlet/airframe/control/Parallel.scala
```diff
@@ -46,17 +65,27 @@ object Parallel {
...
+      val worker = new IndexedWorker[T, R](executionId, i.toString, requestQueue, resultArray, f)
       requestQueue.put(worker)
     }
 
     val executor = Executors.newFixedThreadPool(parallelism)
+    val counter = new AtomicLong(0)
+
+    jmxAgent.foreach { jmxAgent =>
+      val stats = new ParallelExecutionStats(executionId, parallelism, source.size.toString, requestQueue, counter)
+      jmxAgent.register(s"wvlet.airframe.control.Parallel:name=$executionId", stats)
```
xerial: This might register too many mBeans to JVM, and will cause OOM if we frequently use Parallel.run. I think we should use a shared mBean.

airframe-control/src/main/scala/wvlet/airframe/control/Parallel.scala
```diff
@@ -13,16 +13,35 @@
...
-object Parallel {
+object Parallel extends LogSupport {
+
+  class ParallelExecutionStats(
+      @JMX
+      val executionId: String,
+      @JMX
+      val parallelism: Int,
+      @JMX
+      val total: String,
```
xerial: What does `total` mean here? Adding @JMX(description = "...") will be helpful. This description will be shown in jconsole too.

airframe-control/src/main/scala/wvlet/airframe/control/Parallel.scala
xerial: FYI. Datadog has a limitation to the number of metrics (mBeans * its parameters)we can monitor at the same time. It's around several hundreds. 

So we should report fewer metrics for this monitoring:
- How many parallel tasks are running in total?
- How many parallel workers are created in total?

xerial: 350 is Datadog's JMX metrics limit: https://docs.datadoghq.com/integrations/java/#the-350-metric-limit

airframe-control/src/main/scala/wvlet/airframe/control/Parallel.scala
```diff
@@ -13,16 +13,35 @@
...
+
 import scala.reflect.ClassTag
 
 /**
   * Utilities for parallel execution.
   */
-object Parallel {
+object Parallel extends LogSupport {
+
+  class ParallelExecutionStats(
```
xerial: We can create a static instance of this stats, using AtomicInteger, AtomicReference for updating metrics from multiple threads. Then expose such metrics with `@JMX(description = "total worker threads in Parallel") def totalWorkerThreads: Int = AtomicInteger.get()`

airframe-control/src/main/scala/wvlet/airframe/control/Parallel.scala
takezoe: Current `Parallel` implementation is designed for running heavy tasks in parallel, not for running many parallel execution at the same. In this situation, grabbing stats of each execution accurately is better than aggregating them.

But I know that adding / removing metrics dynamically makes monitoring difficult. I would like to allow to be able to be monitored progress of each execution somehow.

xerial:
@takezoe By the way, I just have noticed Presto also has a similar module with Parallel and JMX metrics https://github.com/airlift/airlift/tree/master/concurrent/src/main/java/io/airlift/concurrent

airframe-control/src/main/scala/wvlet/airframe/control/Parallel.scala
takezoe: I researched JMX but I couldn't find any solution to expose metrics per execution as single MBean instance other than returning them as a structured String value like JSON. 😨 

I think aggregated metrics are not enough for current intended use cases (run heavy, but not so many elements in each execution). Do you have any idea?

xerial: - We can use nested mBean. I think you forget to add @JMX annotation to `ParallelExecutionStats` class.

- Then I think JMX registration should be optional (instead of registering each time when `parallel` is used), and we should be able to specify MBean name if JMX monitoring is necessary. Let's talk about it later.

takezoe: I gave up to expose per execution metrics on JMX. Finally, updated to expose aggregated metrics by single static MBean instance. Also MBean registration is now optional.

xerial:
@takezoe Let me know if this PR becomes ready to use.

takezoe:
Got it. I would like to enhance tests for MBean attributes additionally. I will notify you after that.

takezoe:
@xerial I tested JMX metrics using jconsole by hand and added assertions for accumulated attributes to unit tests. Now it's ready to use.

Issue #472, opened 2019-03-28
airframe-http: Add a basic http client


xerial:

xerial:
done in #482

Issue #467, opened 2019-03-27
airframe-control: Remove context from Control.Retrier


xerial:
Passing the context information to Retrier turns out to be a bad design in terms of the usability. We should remove that to favor the simplicity.

xerial:
Done as a part of #482

Issue #366, opened 2019-01-22
Add a basic retry pattern set for HTTP requests


xerial:

xerial:
Done in #482

PR #482, opened 2019-03-31
airframe-http: Add HttpClient

xerial:
Add a common interface and finagle-based implementation for an http client. #472 
- [x] Simplify Retryer (=> RetryContext) to support http error handling #467 #491 
- [x] Add HttpStatus code representations using type classes (xxxAdapter)
- [x] Add a standard HTTP response hander #366 
- [x] Add a finagle-based HTTP client
- [x] Add async/sync http clients. Using F[_] as an abstraction of Future. This is useful to remove the dependency to Finagle in airframe-http.
- [x] Support circuit breaker (No code for this. We can use circuit breaker of Finagle by customizing FinagleClientConfig.initClient)
- [x] Integrate control.Retry and FinagleRetryFilter. 
   - Finagle will not enable request retry by default, so we anyway need to add our own retry filter. 
   - Simplified control.Retry so that we can use this retry logic both for imperative programming and Future-based programming.
- [x] Support response/request -> object mapping with airframe-codec.


airframe-http-finagle/src/main/scala/wvlet/airframe/http/finagle/FinagleRetryFilter.scala
```diff
@@ -0,0 +1,95 @@
...
+                }
+              }
+          } else {
+            // No more retry.
+            x match {
+              case Throw(e) =>
+                warn(e)
+                // Create an error response
+                val r = Response(Status.BadRequest)
+                r.setContentString(cause.getMessage)
```
takezoe: Currently, it looks that there is no way to handle this error message because `HttpClient` has only methods to treat response as JSON. Probably a decoding error happens in this case?

airframe-http-finagle/src/main/scala/wvlet/airframe/http/finagle/FinagleClient.scala
```diff
@@ -0,0 +1,145 @@
...
+  def defaultInitClient: Http.Client => Http.Client = { x: Http.Client =>
+    x.withSessionQualifier.noFailureAccrual
+  }
+  def defaultRetry: RetryContext = {
+    HttpClient.defaultHttpClientRetry[http.Request, http.Response]
+  }
+
+  def newClient(hostAndPort: String, config: FinagleClientConfig = defaultConfig): FinagleClient = {
+    new FinagleClient(address = ServerAddress(hostAndPort), config)
+  }
```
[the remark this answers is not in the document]
xerial: Good point. I've added a fix 3562e82 to automatically add TLS/SSL support for https connections.

airframe-http-finagle/src/main/scala/wvlet/airframe/http/finagle/FinagleRetryFilter.scala
xerial: That's true. This error response will be decoded as a regular response and will through another exception. Let me think about what is a good manner to handle such errors.

xerial: Changed the logic to throw an exception 2b52331

xerial:
CI is failing because of jdk11 download failure. Other than that, we are good to merge this PR.

takezoe:
@xerial Thanks for your quick fix!

PR #484, opened 2019-04-03
airframe-codec: Support JSONValue, Json (raw json string) 

xerial:


airframe-msgpack/src/main/scala/wvlet/airframe/msgpack/spi/Value.scala
```diff
@@ -212,7 +211,7 @@ object Value {
 
   case class TimestampValue(v: Instant) extends Value {
     override def toJson: String = {
-      LogTimestampFormatter.formatTimestamp(v.toEpochMilli)
+      v.toString
```
xerial: We need to support nanosec

PR #498, opened 2019-05-19
Build for Scala 2.11 only at mater branch commits

xerial:
Spark 2.4.x is rolled back to Scala 2.11 because there are many Spark users who still need time to migrate to Scala 2.12. 

This PR will recover Scala 2.11 build only for master branch commits.


xerial:
cc: @takezoe

PR #505, opened 2019-05-31
airframe-metrics: Support exact start dates in TimeWindow

xerial:
If we support extact start dates in TimeWindow, we don't need to support multiple flavors of time window functions (e.g., withinTimeRange(start, end, ...)) Just using `within(duration string)` will be sufficient for the most of the use cases.


airframe-metrics/jvm/src/main/scala/wvlet/airframe/metrics/TimeParser.scala
```diff
@@ -33,29 +33,31 @@ object TimeParser extends LogSupport {
...
           }
       }
       .toOption
   }
 
-  def parseZonedDateTime(s: String): Option[ZonedDateTime] = {
-    def loop(lst: List[DateTimeFormatter]): Option[ZonedDateTime] = {
-      if (lst.isEmpty)
+  def parseZonedDateTime(s: String): Option[TimeParseResult] = {
+    def loop(lst: List[DateTimeFormatter]): Option[TimeParseResult] = {
```
takezoe: We can mark this internal `loop` method as `@tailrec`.

Issue #466, opened 2019-03-26
Exception-free unpackXXX


xerial:
- MessageCodec[A].unpackJson(json:String): Option[A] returns Option[A], but in writing code we need to focus on the logic for using the result A of reading JSON.

- Can we return A without caring about Exception handling? 
- What should be the default error handling if we cannot parse the input json as A type?

xerial:
We added 
- toJson (can throw an exception)
- unpackJson (exception-free)

Similarly:
- toMsgpack (can throw an exception)
- unpackMsgpack (exception-free)

PR #515, opened 2019-06-14
Add convenient methods to convert JSON <-> Scala object

takezoe:
Add convenient methods for JSON conversion to `JSONCodec`.

At first, I thought to add these methods to `aiframe-json`. However, since `airframe-codec` depends on `airframe-json`, adding these methods to `airframe-json` side causes a cyclic dependency between `airframe-json` and `airframe-codec`. So I decided to add these methods to `airframe-codec` side.

Do you have a better idea?

### DIFF
--- a/airframe-codec/src/main/scala/wvlet/airframe/codec/MessageCodec.scala
+++ b/airframe-codec/src/main/scala/wvlet/airframe/codec/MessageCodec.scala
@@ -20,6 +20,7 @@ import wvlet.log.LogSupport
 
 import scala.language.experimental.macros
 import scala.util.{Failure, Success, Try}
+import scala.reflect.runtime.universe._
 
 trait MessageCodec[A] extends LogSupport {
 
@@ -134,4 +135,12 @@ trait MessageValueCodec[A] extends MessageCodec[A] {
 object MessageCodec {
   def of[A]: MessageCodec[A] = macro CodecMacros.codecOf[A]
   def ofSurface(s: Surface): MessageCodec[_] = MessageCodecFactory.defaultFactory.ofSurface(s)
+
+  def fromJson[T: TypeTag](json: String): T = {
+    ofSurface(Surface.of[T]).asInstanceOf[MessageCodec[T]].unpackJson(json).get
+  }
+
+  def toJson[T: TypeTag](obj: T): String = {
+    ofSurface(Surface.of[T]).asInstanceOf[MessageCodec[T]].toJson(obj)
+  }
 }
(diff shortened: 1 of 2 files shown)


xerial:
Adding this feature to airframe-codec is OK, especially because object-mapping should be the out of the scope of airframe-json (just a parser and json model classes).

After merging this PR, we need to consider the error handling of such mappings:
https://github.com/wvlet/airframe/issues/466

I felt returning Option[X] for unpackJson(json) is a little bit inconvenient as we always need to check whether the result is Some or None. 

I think redesigning API like would be better:
```
unpackJson(json):  X
unpackJsonOpt(json): Option[X]
```

xerial:
@takezoe By the way, MessageCodec already has unpackJson, so this enhancement might be confusing, so we should use another object other than JSONCodec.

JSONXXX ?

xerial:
An idea is using MessgeCodec object, for example:
MessageCodec.unpackJson[X](json)

takezoe:
Sounds good. I will move them to MessageCode. Also will take a look at the cause of CI failure.

airframe-codec/src/main/scala/wvlet/airframe/codec/MessageCodec.scala
```diff
@@ -108,6 +109,15 @@ trait MessageCodec[A] extends LogSupport {
         None
     }
   }
+
+  def fromJson[T: TypeTag](json: String): Option[T] = {
```
xerial: Can't we use `object MessageCodec`?

takezoe: Right. Why I added to the class...

takezoe:
Scala 2.13.0 build is still failing in Travis.

By the way, what type of exception should `MessageCodec.fromJson[X:TypeTag](json:String): X` return if conversion fails? Also `Either[Exception, X]` might be better than `Option[T]` in general because we can know the reason of conversion failure.

xerial:
Returning Either means that codec needs to handle exceptions within the codec. But we don't have any standard on how to handle such pack/unpack errors. So simply returning an exception is preferred for the usability.

We can throw `MessageCodecException(INVALID_DATA, codec, error message)`.

takezoe:
I tested JSON conversion using the current airframe-codec with some error patterns. It threw exceptions in some cases. On the other hand, even if there are missing properties or data type difference, any error didn't happen. As you say, we need the standard in error handling if we offer `Either` interface.

My conclusion is that simply calling `Option.get()` to get the content of `Option[X]` is the best solution for now.

xerial:
Moved to #519 to apply some fixes.

PR #519, opened 2019-06-18
Add convenient methods to convert JSON <-> Scala object

xerial:
Move to/fromJson to MessageCodecFactory to use custom codecs


xerial:
@takezoe Moved the implementation to the MessageCodecFactory side to use custom codecs.

PR #529, opened 2019-07-02
Programmable HTTP mock server

takezoe:


airframe-http-recorder/src/main/scala/wvlet/airframe/http/recorder/HttpRecordStore.scala
```diff
@@ -93,21 +93,21 @@ class HttpRecordStore(val recorderConfig: HttpRecorderConfig, dropSession: Boole
...
+        val headerHash = headers.map { x =>
           s"${x._1}:${x._2}".hashCode
-        }
-        .reduce { (xor, next) =>
+        }.reduce { (xor, next) =>
           xor ^ next // Take XOR to compute order-insensitive hash values.
         }
-    prefix.hashCode * 13 + headerHash
+        prefix.hashCode * 13 + headerHash
+    }
```
takezoe: Computing hash value hasn't worked if HeaderMap is empty.

airframe-http-recorder/src/main/scala/wvlet/airframe/http/recorder/HttpRecorder.scala
```diff
@@ -99,6 +97,23 @@ object HttpRecorder extends LogSupport {
...
+    * If no matching record is found, use the given fallback handler.
+    */
+  def createProgrammableServer(programmer: HttpRecordStore => Unit): FinagleServer = {
+    val recorderConfig = HttpRecorderConfig("localhost")
+    val finagleConfig = FinagleServerConfig(recorderConfig.serverPort)
+    val recorder      = new HttpRecordStore(recorderConfig, true){
+      override def requestHash(request: Request): Int = {
+        val content = request.getContentString()
+        s"${request.method.toString()}:${recorderConfig.destAddress.hostAndPort}${request.uri}:${content.hashCode}".hashCode
+      }
```
takezoe: It's difficult to program exact same headers. I think matching by method, address, uri and content is good enough in the programmable HTTP mock server.

airframe-http-recorder/src/main/scala/wvlet/airframe/http/recorder/HttpRecordStore.scala
xerial: Nice catch!

airframe-http-recorder/src/main/scala/wvlet/airframe/http/recorder/HttpRecorder.scala
xerial: ok. Let's add a customization entry point to HttpRecordStore later to add user-defined pattern matching rules.

No change is required for now.

xerial: My idea is adding a method like:
```
HttpRecordStore.withRequestMatcher(matcher:PartialFunction[Request, Response])
```

PR #538, opened 2019-07-06
DSL for JSON extraction

takezoe:
Added following methods:
- `JSONObject.get()`
- `JSONArray.apply()`

Also added JSON extraction DSL for ease:
```scala
val json = """{"user": [{ "id": 1 }, { "id": 2 }]}"""
val id = (JSON.parse(json) / "user" / "id")(0).value.asInstanceOf[Long] // => 1
```

### DIFF
--- a/airframe-json/src/main/scala/wvlet/airframe/json/JSON.scala
+++ b/airframe-json/src/main/scala/wvlet/airframe/json/JSON.scala
@@ -118,6 +118,12 @@ object JSON extends LogSupport {
       s.append("}")
       s.result()
     }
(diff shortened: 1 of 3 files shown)


xerial:
Thanks! 

Honestly speaking I don’t like `\` syntax. Can’t we use `/` instead? It can be confusing with the string division operator, but sbt is already using `/` for file paths without causing such a confusion. So I guess it should be possible to use `/` for JSONValue.

takezoe:
I got you. Updated to use `/` instead.

xerial:
Looks good. Merged

PR #539, opened 2019-07-07
Seq[JSONValue].value

takezoe:
Another `.value` method for JSON DSL added in #538 that allows getting a value of the first element of `Seq[JSONValue]`.

This is useful when we know that the json value has a single child element.

```scala
// before
val id = (jsonValue / "user" / "id")(0).value
// after
val id = (jsonValue / "user" / "id").value
```


xerial:
LGTM!

PR #541, opened 2019-07-09
Use map type MessagePack for ObjectCodec

Lewuathe:
Currently, MessageCodec serializes the value as array type as default even for ObjectCodec which is supporting `PackAsMapSupport`. It can cause a problem of the cooperability between td-agent and airframe-fluentd because td-agent expects the input record to be map type MessagePack. 

Since `TypedMetricLogger` in airframe-fluentd calls `toMsgPack`, the data is flushed as array type. But it cannot be read by td-agent or fluentd. 

```scala
class TypedMetricLogger[T <: TaggedMetric](fluentdClient: MetricLogger, codec: MessageCodec[T]) {
  def emit(event: T): Unit = {
    fluentdClient.emitMsgPack(event.metricTag, codec.toMsgPack(event))
  }
}
```

As we do for `toJson`, `toMsgPack` can write the value as map type if the codec is `PackAsMapSupport`. 

Additionally, using object map type codec in TypedMetricLogger might be good to resolve the cooperability with td-agent.

### DIFF
--- a/airframe-fluentd/src/main/scala/wvlet/airframe/fluentd/MetricLogger.scala
+++ b/airframe-fluentd/src/main/scala/wvlet/airframe/fluentd/MetricLogger.scala
@@ -15,7 +15,7 @@ package wvlet.airframe.fluentd
 import java.util.concurrent.ConcurrentHashMap
 
 import javax.annotation.PreDestroy
-import wvlet.airframe.codec.MessageCodec
+import wvlet.airframe.codec.{MessageCodec, MessageCodecFactory}
 import wvlet.airframe.surface.Surface
 import wvlet.log.LogSupport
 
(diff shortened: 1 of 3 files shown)


airframe-fluentd/src/main/scala/wvlet/airframe/fluentd/MetricLogger.scala
```diff
@@ -65,15 +65,17 @@ class MetricLoggerFactory(fluentdClient: MetricLogger) extends LogSupport {
   def getTypedLogger[T <: TaggedMetric: ru.TypeTag]: TypedMetricLogger[T] = {
     loggerCache
       .getOrElseUpdate(Surface.of[T], {
-        val codec = MessageCodec.of[T]
+        // Ensure to serialize as map type of MessagePack
+        val codec = MessageCodecFactory.defaultFactory.withObjectMapCodec.of[T]
```
xerial: To make it customizable, pass this codec as a constructor argument:
```
class MetricLoggerFactory(fluentdClient:MetricLogger, codecFactory: MessageCodecFactory = MessageCodecFactory.defaultFactory.withObjectMapCodec)  
```

xerial:
Merged. Thanks!

Issue #528, opened 2019-07-02
Milestone: Airframe 19


xerial:
This ticket is a placeholder for describing the roadmap for Airframe 2019 releases.
- Milestone: Airframe 20 #839 

# Airframe 19
## airframe-http
- [x] Filter support #540
  - e.g., adding authentication filters before the router
- ~[ ] Add MultiMap for http request/response headers~ We usually don't use generic request/response for now
- [ ] Add http logging #399 
- ~[ ] Add JMX metrics like TwitterServer~ -> airframe-http-finagle supports it
- [x] Static content (e.g, html, CSS) provider #843 
- http server backend
  - [x] finagle
  - akka-http 
- [x] Convenient factories to initialize servers. 
   - The current implementation provides designs. Factory methods like HttpServer.newXXXServer(...) will be more useful. => Added FinagleServerFactory.newFinagleServer
- [ ] OpenAPI YAML/JSON generator
  - Add a very simple annotation for OpenAPI. 
- ~[ ] gRPC support (idea)~ No urgency in our use cases
- [ ] giter8 template (@smdmts)

## airframe-http-finagle
- [x] Upgrade to Finagle 19.6.0 #531
- Admin server  (Having an example Admin server would be good)
   - [ ] Configure JMX parameters (e.g., log level)
   - [ ] monitoring logs
   - [ ] monitoring stats
- ~[ ] Factory methods for building Finagle Requests~ (@takezoe) => [FinagleRequestBuilder](https://twitter.github.io/finagle/docs/com/twitter/finagle/http/RequestBuilder.html) can be used 

## airframe-http-recorder
- [x] http support #527 
- [x] programmable http recorder.  #529
  - Recording pairs of http request/response without actual web services
- [x] record/replay proxy #557 
- [x] Support the expiration of recorded responses #559 
- ~Custom http request matcher~ (headerExclude option seems sufficient for now)
- ~Custom http request handler~

## airframe-http-client
- [x] Default http client bindings. It seems we have no strong need to support multiple http clients.
  - ~[ ] jetty-client~
  - [x] finagle http client
  - ~[ ] akka-http client~
  - ~[ ] okhttp client~
  - ~[ ] https://github.com/softwaremill/sttp~
- [x] Allow adding custom headers #531
- [x] Client builder #760 
- [ ] Generate client-side code from a server definition (e.g., sbt-plugin?)
- [ ] Minimize the client binding interface used for HttpException retry patterns
- [x] JSON chuncked encoding support

[...]

PR #545, opened 2019-07-10
add kinds of diff method in TimeWindow

smdmts:
- Calculate diff TimeWindow between start and end.
https://github.com/wvlet/airframe/issues/528#issuecomment-508300291

### DIFF
--- a/airframe-metrics/jvm/src/main/scala/wvlet/airframe/metrics/TimeWindow.scala
+++ b/airframe-metrics/jvm/src/main/scala/wvlet/airframe/metrics/TimeWindow.scala
@@ -94,7 +94,15 @@ case class TimeWindow(start: ZonedDateTime, end: ZonedDateTime) {
 
   def plus(n: Long, unit: ChronoUnit): TimeWindow  = TimeWindow(start.plus(n, unit), end.plus(n, unit))
   def minus(n: Long, unit: ChronoUnit): TimeWindow = plus(-n, unit)
-  def between(unit: ChronoUnit): Long              = unit.between(start, end)
+
+  def howMany(unit: ChronoUnit): Long = unit.between(start, end)
+
+  def minuteDiff: Long = howMany(ChronoUnit.MINUTES)
+  def hourDiff: Long   = howMany(ChronoUnit.HOURS)
+  def dateDiff: Long   = howMany(ChronoUnit.DAYS)
+  def weekDiff: Long   = howMany(ChronoUnit.WEEKS)
+  def monthDiff: Long  = howMany(ChronoUnit.MONTHS)
+  def yearDiff: Long   = howMany(ChronoUnit.YEARS)
 
   def intersectsWith(other: TimeWindow): Boolean = {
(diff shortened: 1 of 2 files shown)


airframe-metrics/jvm/src/main/scala/wvlet/airframe/metrics/TimeWindow.scala
```diff
@@ -92,8 +92,9 @@ case class TimeWindow(start: ZonedDateTime, end: ZonedDateTime) {
     }
   }
 
-  def plus(n: Long, unit: ChronoUnit)  = TimeWindow(start.plus(n, unit), end.plus(n, unit))
-  def minus(n: Long, unit: ChronoUnit) = plus(-n, unit)
+  def plus(n: Long, unit: ChronoUnit): TimeWindow  = TimeWindow(start.plus(n, unit), end.plus(n, unit))
+  def minus(n: Long, unit: ChronoUnit): TimeWindow = plus(-n, unit)
+  def between(unit: ChronoUnit): Long              = unit.between(start, end)
```
xerial: The meaning of `between` is a bit unclear for me. 

How about using this name?
```scala
def howManyOf(unit: ChromeUnits):Long
``` 

or
```scala
def countUnits(unit: ChronoUnit): Long
```

xerial:
```
TimeWindow.howMany(DAYS) 
TimeWindow.howMany(HOURS)
```

smdmts: @xerial Thank your review. I used by ChronoUnit's method name but it's not clear just pointed out and I thought so.

How do you think the name of 'datediff'?  It is usually used by SQL functions method names.

1. `datediff`
2. `howMany`
3. `countUnits `

I like to use 1 or 2.

xerial: Yeah. I know `ChromeUnit.DAYS.between(s, e)`

If we go with datediff, `def diffIn(unit)` will be better as we can use this for various time units.

I think `howMany(unit)` has the most straightforward meaning.

xerial: It is also possible to add these functions for convenience:
```
def dateDiff: Long = howMany(DAYS)
def monthDiff: Long = howMany(MONTHS)
...
```

xerial:
@smdmts I've got your idea, so we can implement the suggested fixes on your behalf.
If you want us to do that, let us know.

airframe-metrics/jvm/src/main/scala/wvlet/airframe/metrics/TimeWindow.scala
smdmts: @xerial Thanks in your advice. I'll fix and implement the idea in next.

xerial:
Merged. Thanks for the wonderful idea :)

PR #549, opened 2019-07-19
fix: GraalVM support

andreaTP:
This is a switch to enable using airframe-log on GraalVM.


xerial:
Interesting. Does aiframe-log work on GraalVM?

xerial:
CI is failing because of the code formatting. Please apply `$ sbt scalafmt`

andreaTP:
fixed the formatting, sorry.

Apparently it works, here you have an example:
https://github.com/andreaTP/airframe-log-graal

Thanks for the awesome library!

xerial:
That's good to know because I've never tried GraalVM by myself.

Merged the fix. Thanks!

PR #552, opened 2019-07-24
airframe-http-client: Add getResource for GET with query strings

xerial:
For passing GET request objects as URL query strings.

This will add two more convenient mappings:
- binding URL query strings (?p1=v1&p2=v2& ...) as method arguments
- binding URL query strings to a method object (ResourceRequest(p1, p2, ...))


xerial:
cc: @takezoe

PR #553, opened 2019-07-25
airframe-http: Support reading raw response as is

xerial:


xerial:
@takezoe This is a minor improvement

PR #564, opened 2019-07-30
Remove JSON BigInt values handling 

smdmts:
A large number could not parse cause it's using Long limitation. Therefore, I support to use BigInt-type if it exceeded Long.MAX or Long.MIN values.

- Appendix, JSON's specification is no limit of numeric amounts.
https://www.json.org/

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -651,7 +651,7 @@ lazy val json =
       description := "JSON parser"
     )
     .jsSettings(jsBuildSettings)
-    .dependsOn(log, airframeSpec % "test")
+    .dependsOn(log, msgpack, airframeSpec % "test")
 
 lazy val jsonJVM = json.jvm
 lazy val jsonJS  = json.js
--- a/build.sbt
+++ b/build.sbt
@@ -755,4 +755,13 @@ lazy val examples =
       libraryDependencies ++= Seq(
         )
     )
-    .dependsOn(codecJVM, airframeJVM, airframeMacrosJVM, launcher, jmx, jdbc, tablet, finagle, airframeSpecJVM % "test")
+    .dependsOn(codecJVM,
+               airframeJVM,
+               config,
+               airframeMacrosJVM,
+               launcher,
+               jmx,
+               jdbc,
+               tablet,
+               finagle,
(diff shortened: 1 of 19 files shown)


xerial:
If we add JSONBigInt, JSONCodec and JSONValueCodec also need to support this type. 

If CI is not failing, I guess some important test cases are missing (e.g., reading big int numbers from JSON).

xerial:
msgpack doesn't support values larger than 2^63-1, so for this bug numbers, we need to encode decimal values as strings, then read them as BigInteger objects. 

I'm not sure how many changes are necessary to support BigInteger types. Let me take a look later.

xerial:
We will need:
- [ ] BigIntegerCodec
- [ ] Throw IntegerOverflowException when reading numbers larger than Long.MaxValue (or smaller than Long.MinValue) when mapping to Long value 
- [ ] JSONCodec support 
- [ ] JSONValueCodec support

xerial:
Design options:
- A) Implement above tasks
- B) Forget about BigInt handling and just throw IntegerOverflowException. This might make sense as JSON is not suited to representing scientific values like double (JSON representation of double values will lose the precision)

smdmts:
@xerial I suppose B is better than A cause there are more problems contains A. 

For instance, 
- Msgpack is not supported by BigInteger values (https://github.com/msgpack/msgpack/issues/206)
- JSON's Integer number value of exceeded Long.MaxValue or less than Long.MinValue is not used by many of realistically an application but used instead of String Values.

Therefore, I will be implementing plan B, please give me your opinion.

xerial:
ok. Let’s go with plan B)

build.sbt
```diff
@@ -651,7 +651,7 @@ lazy val json =
       description := "JSON parser"
     )
     .jsSettings(jsBuildSettings)
-    .dependsOn(log, airframeSpec % "test")
+    .dependsOn(log, msgpack, airframeSpec % "test")
```
smdmts: Is there no problem with this as dependsOn?

xerial: It's OK for now. I'll follow-up this issue in #574

Issue #559, opened 2019-07-29
http-recorder: Support expiration


xerial:
Currently expirationTime config of HttpRecorder is not properly used.

PR #567, opened 2019-07-31
airframe-http-recorder: #559 Support expiration

xerial:


xerial:
@takezoe Implemented a missing feature. For now, this will not expire any records while running the recorder.

PR #569, opened 2019-08-02
airframe-http: Wrap exceptions with Future for each filter and context

xerial:


xerial:
@takezoe This is for supporting Future.rescue for each HttpFilter

PR #570, opened 2019-08-02
airframe-http: Support Router.add[Filter], andThen[Filter]

xerial:
- Use a consistent syntax for Router.add[X] for both controller and filter
- Deprecate Router.filter[Filter] 
- Support chaining filters with Router.andThen[Filter]


xerial:
An improvement for avoiding unnecessary Router nesting.

PR #576, opened 2019-08-05
Upgrade to Scala 2.12.9

xerial:
- [x] Scala.js support https://github.com/scala-js/scala-js/issues/3738
  - It hits a bug of Scala compiler https://github.com/scala/bug/issues/11534 so we probably need to wait Scala 2.12.10, which will have this fix https://github.com/scala/scala/pull/8261

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,7 +41,7 @@ matrix:
       jdk: openjdk8
       script:
         - cd $TRAVIS_BUILD_DIR && ./scripts/publish.sh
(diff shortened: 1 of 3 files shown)


xerial:
It seems we should wait Scala 2.12.10 because of this string interpolation regression https://github.com/scala/bug/issues/11665 in https://github.com/scala/bug/milestone/102. Spark will also wait for 2.12.10 https://github.com/apache/spark/pull/25404

PR #578, opened 2019-08-07
airframe-http-recorder: Support custom recording 

xerial:
- [x] Support custom recorder configs in the programmable server
- [x] Add in-memory programmable server for debugging purpose
- [x] Support recording custom request/response pairs in HttpRecorderServer

@takezoe This PR enables combining real server responses + programmed responses, which will be useful for simulating server failures.

PR #579, opened 2019-08-07
airframe-http-recorder: Support binary requests and responses

xerial:
@takezoe This will be useful to record binary responses (e.g., from S3)

### DIFF
--- a/airframe-http-recorder/src/main/scala/wvlet/airframe/http/recorder/HttpRecord.scala
+++ b/airframe-http-recorder/src/main/scala/wvlet/airframe/http/recorder/HttpRecord.scala
@@ -16,7 +16,7 @@ import java.nio.charset.StandardCharsets
 import java.sql.{Connection, ResultSet}
 import java.time.Instant
 
-import com.twitter.finagle.http.{Response, Status, Version}
+import com.twitter.finagle.http.{MediaType, Request, Response, Status, Version}
 import com.twitter.io.Buf
 import wvlet.airframe.codec._
 import wvlet.airframe.control.Control.withResource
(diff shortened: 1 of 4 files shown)


airframe-http-recorder/src/test/scala/wvlet/airframe/http/recorder/HttpRecorderTest.scala
```diff
@@ -170,4 +173,36 @@ class HttpRecorderTest extends AirframeSpec {
     }
   }
 
+  "support binary contents" in {
+    val storeConfig = HttpRecorderConfig(destUri = "localhost", sessionName = "binary-test")
+    val store       = new HttpRecordStore(storeConfig, dropSession = true)
```
xerial: I'll use the in-memory recorder after #578 is merged.

airframe-http-recorder/src/main/scala/wvlet/airframe/http/recorder/HttpRecord.scala
```diff
@@ -43,13 +43,23 @@ case class HttpRecord(session: String,
...
-    r.content = Buf.ByteArray.Owned(rawBytes)
-    r.contentLength = rawBytes.length
+    val r = Response(Version.Http11, Status.fromCode(responseCode))
+
     responseHeader.foreach { x =>
       r.headerMap.set(x._1, x._2)
     }
+
+    val contentBytes = responseHeader.find(h => h._1.equalsIgnoreCase("content-type")) match {
+      case Some(x) if x._2 == MediaType.OctetStream =>
```
takezoe: Supporting only octet-stream is OK? Rather, treating all responses as binary might be better?

xerial:
@takezoe That's a good idea. To do so, we need to discard the previously recorded results as it will break the compatibility.

PR #580, opened 2019-08-07
AirSpec: A new testing framework for Scala

### LINKED ISSUE #581: sbt: Clean-up java.util.logging LogHandlers to avoid ClassNotFound issue
While developing #580, I've found sbt 1.3.0 is not cleaning up log handlers registered in java.util.logging. If project classes in wvlet.log is used in a handler, java.util.logging fails to read such classes after discarding the previous class loader. 

As a workaround, we always need to clean-up log handlers registered to java.util.logging.

xerial:
image

Basic ideas:
- Extends AirSpec trait to define test cases
- Public methods in AirSpec will be test cases
- Function arguments will be injected by Airframe DI. Session will be created/discarded per AirSpec class => configure(Design):Design, configureLocal(Design):Design

Considerations:
- Should we run all public methods? => Yes
- How to add a test description to each test function? => Use `symbol...` method names
- Scala.js support (How to embed Surface in Scala.js, which has no runtime reflection?)
  - => Resolved: Users need to implement `AirSpec.methodSurface = Surface.methodSurfaceOf[X]`. We may be able to simplify this by using Scala annotation macro in future.  => Added scalaJsSupport method

Tasks:
- [x] Runner
- [x] AirSpec base trait for DI based testing
- [x] Read test methods from methodSurfaces
- [x] Scala.js cross-build
- [x] Fixed #581 (ClassNotFound LogLevel$)
- [x] Reporting test results
- [x] Selecting test methods using command-line options
- ~[ ] Add simple test("....") spec like minitest~
- [x] basic usage README


xerial:
The first implementation that supports minimum requirements is done. 

- Rename airframe-spec -> airframe-scalatest for avoiding naming confusion
- DI-based testing is working, but it seems we need more experiments by actually writing tests with AirSpec.
- Other enhancement (e.g., power assertion, integration with ScalaCheck) can be done later.

cc: @takezoe @Lewuathe

PR #619, opened 2019-08-20
airframe-http-recorder: Store records to .airframe/http folder

xerial:
- And allow configuring in-memory recorders
- Use 1M default expiration range

This change is for not creating too many database files.


airframe-http-recorder/src/main/scala/wvlet/airframe/http/recorder/HttpRecordStore.scala
```diff
@@ -154,7 +156,7 @@ class HttpRecordStore(val recorderConfig: HttpRecorderConfig, dropSession: Boole
       createdAt = Instant.now()
     )
 
-    trace(s"Recording ${request} -> ${entry.summary}")
+    trace(s"record: request hash ${rh} for ${request} -> ${entry.summary}")
```
xerial: Show the request hash for the ease of debugging (cc: @takezoe )

PR #643, opened 2019-08-28
Add a stream json-msgpack converter

xerial:
This PR adds an efficient conversion method from JSON to msgpack

cc: @takezoe


xerial:
@takezoe I'll merge it as it just adds more functionality. We may need to add more helper methods to support InputStream, Array[Byte] if we are going to use this feature.

Issue #644, opened 2019-08-28
airframe-di: Close resources implementing AutoClosable at shutdown


xerial:
If we automatically close resources that implement AutoClosable at session shutdown, we can avoid writing onShutdown hooks or `@PreDestroy` annotations at all.

```scala
class Test extends AirSpec {
   protected override def design = newDesign
     .bind[HttpClient].toInstance(new HttpClient(...))  // Assume HttpClient is extending AutoClosable
 
   def test(client:HttpClient): Unit = {
   }
   // HttpClient will be closed automatically after the spec fiinishes
}
```

To avoid conflicts with multiple shutdown hooks defined in a class, we may need to define the precedence to avoid duplicate shutdown hook calls. For example, 
1. onShutdown hooks, methods with `@PreDestory` (the same rank)
1. AutoClosable.close() (call it if no onShutdownHook nor `@PreDestory` annotation is found)

xerial:
Done in #645

PR #645, opened 2019-08-29
airframe-di: Automatically register shutdown hooks for Closeable resources

xerial:
#644 

@takezoe @Lewuathe 
What I'm expecting for this review is getting questions around how Airframe DI's lifecycle hooks work in general, rather than checking the details of this PR. 

Related code:
- default life cycle handler chain: https://github.com/wvlet/airframe/blob/b20fd4e98ea51817477b2d49319d4922472b26e1/airframe/src/main/scala/wvlet/airframe/LifeCycleManager.scala#L181

- init life cycle hook is called when a new instance is injected:  https://github.com/wvlet/airframe/blob/b20fd4e98ea51817477b2d49319d4922472b26e1/airframe/src/main/scala/wvlet/airframe/AirframeSession.scala#L197

### DIFF
--- a/airframe/src/main/scala/wvlet/airframe/LifeCycleManager.scala
+++ b/airframe/src/main/scala/wvlet/airframe/LifeCycleManager.scala
@@ -15,6 +15,7 @@ package wvlet.airframe
 
 import java.util.concurrent.atomic.AtomicReference
 
+import wvlet.airframe.CloseableLifeCycleHookFinder.CloseHook
 import wvlet.airframe.surface.Surface
 import wvlet.airframe.tracing.Tracer
 import wvlet.log.{LogSupport, Logger}
@@ -92,17 +93,23 @@ class LifeCycleManager(
   def preShutdownHooks: Seq[LifeCycleHook] = preShutdownHookHolder.list
   def shutdownHooks: Seq[LifeCycleHook]    = shutdownHookHolder.list
 
-  protected def addHook(h: LifeCycleHook)(body: LifeCycleManager => Unit): Unit = {
+  protected def findLifeCycleManagerFor[U](s: Surface)(body: LifeCycleManager => U): U = {
     // Adding a lifecycle hook in the owner session
-    session.findOwnerSessionOf(h.surface) match {
+    session.findOwnerSessionOf(s) match {
       case Some(s) => body(s.lifeCycleManager)
       case None    => body(this)
     }
   }
 
+  private[airframe] def hasShutdownHooksFor(s: Surface): Boolean = {
+    findLifeCycleManagerFor(s) { l =>
+      l.shutdownHookHolder.hasHooksFor(s)
+    }
+  }
+
   def addInitHook(h: LifeCycleHook): Unit = {
-    addHook(h) { l =>
-      if (l.initHookHolder.isFirstRegistration(h)) {
+    findLifeCycleManagerFor(h.surface) { l =>
+      if (l.initHookHolder.registerOnlyOnce(h)) {
         debug(s"[${l.sessionName}] Add an init hook: ${h.surface}")
         h.execute
       } else {
@@ -112,7 +119,7 @@ class LifeCycleManager(
   }
 
   def addInjectHook(h: LifeCycleHook): Unit = {
-    addHook(h) { l =>
+    findLifeCycleManagerFor(h.surface) { l =>
       debug(s"[${l.sessionName}] Running an inject hook: ${h.surface}")
       // Run immediately
       h.execute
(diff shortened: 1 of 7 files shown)


airframe/src/test/scala/wvlet/airframe/CloseableShutdownHookTest.scala
```diff
@@ -0,0 +1,59 @@
...
+
+import wvlet.airspec.AirSpec
+
+/**
+  *
+  */
+class CloseableShutdownHookTest extends AirSpec {
+  scalaJsSupport
+
+  class A extends AutoCloseable {
```
xerial: @takezoe Note. with the support for inner classes in Surface, we can define a test class closer in a closer location to the test method.

airframe/src/test/scala/wvlet/airframe/CloseableShutdownHookTest.scala
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    a.closeIsCalled.get() shouldBe true
+  }
+
+  trait A1 {
+    val onShutdownIsCalled = new AtomicBoolean(false)
+    val a = bind[A].onShutdown { x =>
+      onShutdownIsCalled.set(true)
+    }
```
takezoe: How does the shutdown process work if a singleton component is bound in multiple places with `onShutdown` hook? For example:

```scala
trait A1 {
  val a = bind[A].onShutdown { ... }
}

trait A2 {
  val a = bind[A].onShutdown { ... }
}

val design = newDesign
  .bind[A].toSingleton // or toInstance(), or simply bind[A]?
```

airframe/src/main/scala/wvlet/airframe/LifeCycleManager.scala
```diff
@@ -256,3 +295,30 @@ object FILOLifeCycleHookExecutor extends LifeCycleEventHandler with LogSupport {
...
+  * close() if there is no other shutdown hooks
+  */
+object CloseableLifeCycleHookFinder extends LifeCycleEventHandler with LogSupport {
+
+  class CloseHook(val injectee: Injectee) extends LifeCycleHook {
+    override def toString: String = s"CloseHook for [${surface}]"
+    override def execute: Unit = {
+      injectee.injectee match {
+        case c: AutoCloseable =>
+          c.close()
```
takezoe: If an exception occurs in lifecycle hooks, how does Airframe work?

airframe/src/test/scala/wvlet/airframe/CloseableShutdownHookTest.scala
xerial: If A is a singleton, only the first registered shutdown hook will be used.

For example, if A1 is created first, its onShutdown hook will be used and A2's shutdown hook will be ignored.

airframe/src/main/scala/wvlet/airframe/LifeCycleManager.scala
xerial: For now, it will just throw the exception at here:
https://github.com/wvlet/airframe/blob/4112e950fd48072213491abfc329bba344d18ebc/airframe/src/main/scala/wvlet/airframe/LifeCycleManager.scala#L81

So the other shutdown hooks will not be called. It would be good to add an error handler so that we can customize how to handle such exceptions.

airframe/src/test/scala/wvlet/airframe/CloseableShutdownHookTest.scala
xerial: I usually feel difficulties to find an appropriate place to write such shutdown hooks. As in your examples, we may define duplicate shutdown hooks and its behavior can be confusing, but at the same time, we should not call shutdown hooks multiple times. 

So the purpose of this PR is for using a pre-defined resource deallocation method (close()) if exists. I think this will reduce the needs for defining onShutdown hooks.

airframe/src/main/scala/wvlet/airframe/LifeCycleManager.scala
xerial: Added a new ticket for handling exceptions #649

airframe/src/test/scala/wvlet/airframe/CloseableShutdownHookTest.scala
takezoe: Yes, I agree with difficulties in defining shutdown hooks at binding-time. But a way to define shutdown hook is still necessary because making all components which need shutdown process is no reality. So I think we would be better to deprecate the binding-time shutdown hook definition and encourage to use the defining-time shutdown hook definition.

xerial: I also think adding shutdown hooks to a design is a good idea (I sometimes want to do that). I’ll file a ticket

PR #653, opened 2019-09-02
airframe-di: throw exceptions occurred at shutdown hooks later

### LINKED ISSUE #649: airframe-di: Exception safe shutdown hook chain
For now, if some exception happens during an onShutdown hook process, the other subsequent shutdown hooks will not be called. 

To properly close resources, we should handle exceptions at FILOLifeCycleExecutor

xerial:
Resolves #649

### DIFF
--- a/airframe-di-macros/src/main/scala/wvlet/airframe/AirframeMacros.scala
+++ b/airframe-di-macros/src/main/scala/wvlet/airframe/AirframeMacros.scala
@@ -46,7 +46,7 @@ private[wvlet] object AirframeMacros {
         if (t.toString.contains("#")) {
           false
         } else {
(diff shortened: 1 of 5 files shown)


airframe-di-macros/src/main/scala/wvlet/airframe/AirframeMacros.scala
```diff
@@ -46,7 +46,7 @@ private[wvlet] object AirframeMacros {
         if (t.toString.contains("#")) {
           false
         } else {
-          true
+          hasPublicDefaultConstructor
```
xerial: We can't instantiate inner classes without any public default constructor

PR #685, opened 2019-09-09
[WIP] Airspec launcher

xerial:
Create a standalone launcher for AirSpec


xerial:
We need to redo this work as we added new test syntax #876

PR #687, opened 2019-09-09
[Test] Add CircleCI build settings

xerial:


xerial:
It seems to make it work we need to overcome several challenges. It will be a waste of time for now. I'm going to close this.

PR #731, opened 2019-09-18
Refactor DesignOptions using copy method

kmizu:
Since `DesignOptions` is case class, the refactored code using  `copy` method is more concise.


xerial:
Merged. Thanks!

PR #732, opened 2019-09-18
Upgrade to Scala.js 1.0.0-M8

xerial:


build.sbt
```diff
@@ -519,21 +519,17 @@ lazy val metricsJVM = metrics.jvm
 lazy val metricsJS  = metrics.js
 
 lazy val airframeScalaTest =
-  crossProject(JVMPlatform, JSPlatform)
+  project
```
xerial: Drop ScalaTest support for Scala.js. AirSpec can be used instead.

PR #748, opened 2019-10-01
Add unidoc test

xerial:


xerial:
Removed unidoc task as having a single scala doc is not so useful (because sonatype API is quite slow recently)

PR #757, opened 2019-10-03
MetricLogger shouldn't throw exception

takezoe:
Fluency throws an exception when the buffer is full and fail to flush, but it shouldn't break the host logic even if such unknown errors occur inside of it.

### DIFF
--- a/airframe-fluentd/src/main/scala/wvlet/airframe/fluentd/MetricLogger.scala
+++ b/airframe-fluentd/src/main/scala/wvlet/airframe/fluentd/MetricLogger.scala
@@ -18,6 +18,7 @@ import javax.annotation.{PostConstruct, PreDestroy}
 import wvlet.airframe.codec.{MessageCodec, MessageCodecFactory}
 import wvlet.airframe.surface.Surface
 import wvlet.log.LogSupport
+import scala.util.Try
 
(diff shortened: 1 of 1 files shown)


xerial:
Most likely AnyCodec has an issue if this part is throwing an exception

xerial:
@takezoe So we also need to wrap TypeMetricLogger.emit

takezoe:
I see. I'll update to wrap `emit()`.

xerial:
It would be great if there is an way to show warning messages at least once for the same type of errors. For example, while debugging we usually need to know the cause of errors.

xerial:
Second thought: 
- Essentially we need a circuit breaker if reporting metrics becomes difficult due to fluency or network issues. This feature can be added to airframe-control.
- Usually throwing an exception is better so that the client side program can decide what to do if metrics cannot be send. For example, we can store metrics to a disk, then resume sending metrics to fluentd if it recovers. This behavior should be configurable (e.g., report warning, ignore exceptions, throw exceptions)

takezoe:
Certainly, this is not for only logging but also for sending system metrics. If sending metrics is resposibility of the client, exception should be handled in the client-side. OK, I withdraw this pull request and implement error handling in the client-side.

Issue #758, opened 2019-10-04
airframe-control: Add circuit breaker


xerial:
For handling issues like #757

xerial:
Done in #763

PR #763, opened 2019-10-06
airframe-control: Add CircuitBreaker

xerial:
Implements #758 

## CircuitBreaker States:
- closed 
  - Requests will be processed
- open
  - Throws an exception (fail-fast) to protect the external service
  - Wait until a given delay timeout (backoff), then move to half-open state
- half-open
  - try to send a request 
    - success: move to closed state
    - failure: move to open state

## tasks
- [ ] ~Integration with Retry~
- [x] Use RetryPolicy for backoff delay
- [x] Add the capability to monitor the state of CircuitBreaker to know which external services were not working.
- [x] Integration with airframe-http-client (async based implementation) -> Added stand-alone mode to support this usage

Related work:
- https://doc.akka.io/docs/akka/current/common/circuitbreaker.html
- https://monix.io/docs/2x/eval/circuit-breaker.html
- https://github.com/jhalterman/failsafe/blob/master/src/main/java/net/jodah/failsafe/CircuitBreaker.java
- https://github.com/twitter/finagle/blob/develop/finagle-core/src/main/scala/com/twitter/finagle/liveness/FailureAccrualFactory.scala


xerial:
@takezoe I think this supports the basic functionality of circuit breaker that can be configured to our use case.

xerial:
@takezoe I think creating a snapshot version of this CircuitBreaker is better for testing this. So let me merge it and fix the implementation as we find ideas for improving the design.

PR #773, opened 2019-10-13
No need to shutdown ConnectionPoolFactory explicitly

takezoe:
Because `ConnectionPoolFactory` is `AutoCloseable`.


xerial:
Nice catch.

PR #775, opened 2019-10-15
Fix airframe-config README

zerosum:
I think that `*default` should need an anchor in the example(`server.yaml)`.


xerial:
Thanks. Merged

PR #786, opened 2019-10-26
Deprecate runtime provider binding definitions

takezoe:
Though this is just my opinion, runtime binding definition is confusing. For example, imagine that we have following in-trait injections,

```scala
val conn = bind[Connection]{ config: DatabaseConfig1 =>
  config.newConnection()
}

val conn = bind[Connection]{ config: DatabaseConfig2 =>
  config.newConnection()
}
```

Although both initializers take a different type of argument, only either one is actually called because `Connection` instance is created by the one called first and that instance is reused in other places.

Certainly, we can avoid this situation by giving another type to the component using tagged type or type alias. However, it's hard if we haven't grasped the entire codebase. Unintentional reusing an instance created by other binding definition can happen.

In order to avoid this kind of confusion, we should use design-time binding definition in this case.

### DIFF
--- a/airframe/src/main/scala/wvlet/airframe/package.scala
+++ b/airframe/src/main/scala/wvlet/airframe/package.scala
@@ -48,11 +48,17 @@ package object airframe {
     * @tparam A
     */
   def bind[A]: A = macro bindImpl[A]
+  @deprecated(message = "Use design-time binding", since = "19.10.2")
   def bind[A](provider: => A): A = macro bind0Impl[A]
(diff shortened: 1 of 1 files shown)


xerial:
Now that lifecycle can be designed, we can move this type of initialization code entirely to the design side. We also need to update README.md

xerial:
Updated doc as well : https://github.com/wvlet/airframe/commit/1a669d8e4f46f39030fb7be5dcc7edd41a6d15ea

Issue #804, opened 2019-11-10
airframe-di: Add in-trait bindLocal syntax


xerial:
Adding a support for initializing local instances, whose lifecycle will be managed by Airframe session:
```scala
trait A {
  // Create a new instance of Executor locally when A is initialized
  val threadManager: Executor = bindLocal { Executors.newFixedThereadPool(1) }
}
```
This is convenient when the initialization logic is quite local to the implementation (e.g., RAII: https://en.wikipedia.org/wiki/Resource_acquisition_is_initialization)

By supporting this `bindLocal {... }` syntax, we can simplify airframe DI as follows:
-  In-trait `bind[X]` will no longer support any initialization code (like provider bindings). X needs to be defined elsewhere
- Only `bindLocal { (d1, d2, ...) => X}`  will support having local initialization code block. A new instance of X will be created every time using the initialization code block. It means, we cannot override the implementation defined with this in-trait `bindLocal`. 
   - It's still possible to change d1, d2 in the global design. So the behavior can be configurable in the local initialization code block
- If we need to create a new instance of X every time and need to change the binding between production/test, using in-trait `bind[X]` and `design.bind[X].toInstanceProvider { ... }` will work.

By having this clear separation between `bind[X]` and `bindLocal`, users who prefer pure-DI (enable configuring all bindings externally) can write code by using only `bind[X]`.

xerial:
With this approach the flower-bundle pattern (mix-in) will be unnecessary because we can always bind components locally without using trait mix-in.

takezoe:
Hmm... If we need is lifecycle management, we can use host component's lifecycle methods. I think this covers all cases where `bindLocal` will be used.
```scala
trait A {
  val config = bind[ConnectionPoolConfig]
  var cp: ConnectionPool = null

  @PostConstruct
  def start(): Unit = { cp = new ConnectionPool(config) }

  @PreDestroy
  def stop(): Unit = { cp.close() }
}
```

xerial:
But it needs to use `var`. If we have bindLocal, we can use val, which is much safer in Scala.

xerial:
Another benefit of using such provider bindings is we don’t need to manage the initialization order of dependencies. `bindLocal { (d1, d2, ...) => new X(d1, d2, ...) }` will ensure initializing d1 and d2 first before creating X2. If we use your example style, an extra care should be taken to avoid NPE.

takezoe:
[...]

PR #805, opened 2019-11-10
airframe-di: Add in-trait bindLocal

xerial:
#804


xerial:
Thanks. Merged to the master. I'll create a new release so that we can test this.

PR #816, opened 2019-11-20
airframe-di: Support in-class bind[X] for classes with DISupport trait

xerial:


docs/src/main/tut/docs/internals.md
```diff
@@ -52,53 +52,58 @@ To create instances of `A` and `B` inside `App`, we need to pass the concrete in
...
+In the above code, `A` will be instantiated with DISupport trait, which has `session` definition. `bind[B]` inside trait `A` will be expanded liks this similarly:
+
 ```scala
-new A extends SessionHolder {
+new A extends DISupport {
   // (original code) val b = bind[B]
   val b: B = {
     val session = findSession(this)
-    val binder = session..getOrElseUpdate(Surface.of[B], (session:Session => new B with SessionHolder { ... } ))
+    val binder = session..getOrElseUpdate(Surface.of[B], (session:Session => new B with DISupport { ... } ))
```
shimamoto: double dots `session..getOrElseUpdate` typo?

Just a quick question, does getOrElseUpdate method in wvlet.airframe.Session trait exist?

xerial: Nice catch. I don't remember why I wrote this pseudo code. I'll fix it.

xerial: Fixed the example

PR #831, opened 2019-11-30
Add raw response methods to HttpClient

takezoe:
Add some convenient methods to `HttpClient` for handling a raw response.


xerial:
@takezoe Please add test methods to make sure these methods are working as expected. Other than that, this PR is OK.

takezoe:
Thanks for reviewing. I added tests.

In addition, `deleteOpsRaw` is also useful?
```scala
  def deleteOpsRaw[Resource: ru.TypeTag](
      resourcePath: String,
      resource: Resource,
      requestFilter: Req => Req = identity
  ): F[Resp]
```

xerial:
For deleteOps, we can use `deleteOps[Resource, http.Response]`, so we don't need deleteOpsRaw

xerial:
In this context, xxxRaw[Resource] methods will be just wrappers over `xxxOps[Resource, Response]`.

xerial:
ok. LGTM.

PR #860, opened 2019-12-17
sbt 1.3.5

eed3si9n:
This fixes the crossScalaVersions of `sql` subproject.
Ref https://github.com/wvlet/airframe/commit/b8864d3482c20d7c01e22ae5f3be8519431b0df9
Ref https://github.com/sbt/sbt/issues/5310


xerial:
@eed3si9n Thanks! So it seems sbt-1.3.5 traverses these cross build dependencies in advance even if they are not used in a task.

PR #869, opened 2019-12-23
publish for Scala 2.13 + Scala.Js

andreaTP:
this is not tested ... but is the best attempt I can think about ....


xerial:
It seems we need additional work to make it work. Continuing at #870

PR #876, opened 2020-01-01
airspec: Add test(...) syntax

### LINKED ISSUE #872: airspec: Add test(...) syntax for defining local designs
```scala
class A extends AirSpec {
  test("simple test", design = ...) { (d1:D1, d2:D2, ...) => // Inject dependencies
     ....
  }
}
```

xerial:
This will close #872 

- Support local design update
- Support nesting test cases at ease
- This test method does not require `scalaJsSupport` call for Scala.js
- [x] Update doc

image

image


xerial:
@shimamoto This new syntax makes writing tests in AirSpec convenient especially if you need to use the same http client instance between tests.

PR #880, opened 2020-01-04
[WIP] airspec: Async test support

xerial:
If the return type of a test method is Future[_], we should retrieve the result asynchronously. 

- For JVM, Await.result(Future[_], Duration.Inf) works
- For Scala.js, there is implementing Await.result is not straightforward because JavaScript doesn't have the notion of blocking. 

Approaches to support Async testing in Scala.js would be:
- A) Wait until the Future.value becomes available 
- B) Build a tree of Future results, and run tests asynchronously. 

A) is technically impossible as the waiting context will hold the execution control (so Future will never complete while waiting the completion), so we need to implement B), which is better as we can also use this effort to implement a standalone AirSpec launcher.


xerial:
This will require large refactoring of the current code base. We also need to introduce some functionality to chain Futures. I'll work on this elsewhere

PR #886, opened 2020-01-12
Add option to disable implicit instance creation

takezoe:
Add a new option that disables implicit instance creation.

```scala
val d = Design.newDesign
  .bind[Component].toSingleton
  .noDefaultInstanceInjection
```

This allows us to prevent unexpected side effect caused by unintentional instance creation. Also, it helps us to remove unused components with confidence.

### DIFF
--- a/airframe/src/main/scala/wvlet/airframe/AirframeSession.scala
+++ b/airframe/src/main/scala/wvlet/airframe/AirframeSession.scala
@@ -353,6 +353,11 @@ private[airframe] class AirframeSession(
 
     val result =
       obj.getOrElse {
+        // strict mode
+        if (design.designOptions.disableImplicitInstanceGeneration.contains(true)) {
+          throw new MISSING_DEPENDENCY(seen, sourceCode)
+        }
+
         trace(s"[${name}] No binding is found for ${tpe}. Building the instance. create = ${create}")
         if (create) {
           // Create a new instance for bindFactory[X] or building X using its default value
--- a/airframe/src/main/scala/wvlet/airframe/AirframeSession.scala
+++ b/airframe/src/main/scala/wvlet/airframe/AirframeSession.scala
@@ -355,7 +355,7 @@ private[airframe] class AirframeSession(
       obj.getOrElse {
         // strict mode
         if (design.designOptions.defaultInstanceInjection.contains(false)) {
-          throw new MISSING_DEPENDENCY(seen, sourceCode)
+          throw new MISSING_DEPENDENCY(tpe :: seen, sourceCode)
         }
 
         trace(s"[${name}] No binding is found for ${tpe}. Building the instance. create = ${create}")
--- a/airframe/src/main/scala/wvlet/airframe/AirframeSession.scala
+++ b/airframe/src/main/scala/wvlet/airframe/AirframeSession.scala
@@ -354,7 +354,7 @@ private[airframe] class AirframeSession(
     val result =
       obj.getOrElse {
         // strict mode
-        if (design.designOptions.disableImplicitInstanceGeneration.contains(true)) {
+        if (design.designOptions.defaultInstanceInjection.contains(false)) {
           throw new MISSING_DEPENDENCY(seen, sourceCode)
         }
 
--- a/airframe/src/main/scala/wvlet/airframe/Design.scala
+++ b/airframe/src/main/scala/wvlet/airframe/Design.scala
@@ -29,6 +29,7 @@ import wvlet.airframe.lifecycle.LifeCycleHookType
 case class DesignOptions(
     enabledLifeCycleLogging: Option[Boolean] = None,
     stage: Option[Stage] = None,
+    disableImplicitInstanceGeneration: Option[Boolean] = None,
     options: Map[String, Any] = Map.empty
 ) extends Serializable {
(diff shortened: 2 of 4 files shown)


xerial:
Two points:
- noXXX is the current configuration name convention. 
- What is implicit instance creation? If you mean the behavior of creating new instances even if no design is given for the type, the configuration name should be much easier one. Like, withStrictMode (= always require a binding in the design), or withDesignValidation (and noDesignValidation, etc.)

airframe/src/main/scala/wvlet/airframe/AirframeSession.scala
```diff
@@ -353,6 +353,11 @@ private[airframe] class AirframeSession(
 
     val result =
       obj.getOrElse {
+        // strict mode
+        if (design.designOptions.disableImplicitInstanceGeneration.contains(true)) {
+          throw new MISSING_DEPENDENCY(seen, sourceCode)
```
xerial: Does `seen` contains the current target `tpe`?

takezoe:
>  Like, withStrictMode (= always require a binding in the design), or withDesignValidation (and noDesignValidation, etc.)

Actually, `strictMode` was my first idea. However, I thought that it might be better to have `strictMode` as the combination of `productionMode` and `disableImplicitInstanceCreation` (`noImplicitInstanceCreation`, following the existing naming convention).

> To understand the meaning of the name like implictInstanceCreation users need to know the behavior of Airframe internal

I think that using Airframe without proper internal understanding is terrible because if some components have a side-effect in their constructor and it's instantiated by Airframe unintentionally, it might cause a terrible accident. Actually, I faced this situation some times at work.

Another idea is enabling implicit instance creation for only components marked with annotations like `@Component`.

xerial:
@takezoe One of the design philosophy, which is mentioned in the book, A Philosophy of Software Design, is removing errors out of existence (e.g., missing binding). Providing the reasonable default behavior (creating singletons using the default constructors and parameters) is actually taking the necessary follow-up action on behalf of the users, that is, adding a missed binding (using singleton).

A reason why this philosophy is important is, if we always throw an exception, it will force users to handle exceptions for all possible situations. If the users fail to add such exception handling, it will cause the system crash at unexpected timing. For example, if some binding is used quite rarely and the test case is not covering this corner case, this option will embed a bug that causes MISSING_DEPENDENCY. 

Because of this lacking a safe default behavior, I believe this option should be used only for testing the system, rather than using in production. I think what you really need would be compile-time dependency check, not the runtime exception like this option provides. 

I'm not against introducing this option. But without having compile-time dependency check, this option will not provide the safety what we really need; completing all necessary bindings before deploying the service into production.

xerial:
Throwing a run-time exception for missed binding is not a good idea as we are not perfect (e.g., having 100% code coverage will be difficult), but adding a design validation at startup (while initializing all bindings at production mode) can be used for writing a complete design of the service.

How about adding `withEagerDesignValidation (and noEagerDesignValidation)` (check missed binding while initializing bindings). This option must be used with production mode as in-trait bindings can not be found without initializing instances. 

This will still not cover missed bindings, though.

takezoe:
> Providing the reasonable default behavior (creating singletons using the default constructors and parameters) is actually taking the necessary follow-up action on behalf of the users, that is, adding a missed binding (using singleton).

That's true only if users know that behaviour well and all components have proper default. However, I think it's still difficult to make sure that all auto-instantiated components have proper default even if I have a proper understanding of this behaviour.

I think that a runtime error is better than work with unintentional dependency resolving, but this is just my feeling and also another fact is that registering all components explicitly is bothering. So enabling this kind of safe-guard for only testing might be reasonable for now.

> For example, if some binding is used quite rarely and the test case is not covering this corner case, this option will embed a bug that causes MISSING_DEPENDENCY.

My understanding is that we can detect missing dependencies using the production mode, but I came up with some pitfalls like calling `bind[A]` or `session.build[A]` dynamically. Hmm...

Anyway, validation at startup would be helpful even if it's not a perfect solution.

takezoe:
I think that the automatic singleton binding is not safe default behavior, even though it's convenient since we don't need to register all necessary components to `Design`.

Also, missing dependency error can still happen at unexpected timing [in this case](https://github.com/wvlet/airframe/pull/886/commits/c070d30550733a82d0972d83edc83f6dac288543) even though Airframe provides the automatic singleton binding.

As I mentioned, the automatic singleton binding is quite convenient because it can reduce the amount of necessary binding definitions. However, I think it's not good to consider that it provides safe default behavior.

xerial:
@takezoe 
• It seems your provided example should not throw an exception as the default constructor is available for Config2. It might be a bug in building the default instance (in AirframeSession, or Zero.of[Config2], etc.)
• The safety I mentioned here is a quite limited one; Even if users forgot to add some binding, the system should not crash due to MISSING_DEPENDENCY at runtime. It doesn't necessary mean the application works normally.

xerial:
Naming suggestion: `.noDefaultInstanceInjection`

airframe/src/main/scala/wvlet/airframe/AirframeSession.scala
takezoe: Thanks! Fixed.

PR #894, opened 2020-01-15
Update CircuitBreaker

takezoe:
- `run()` returns the result value
- Introduce `onFailure` handler
- Introduce `CircuitBreakerRecoveryPolicy`

### DIFF
--- a/airframe-control/src/main/scala/wvlet/airframe/control/CircuitBreaker.scala
+++ b/airframe-control/src/main/scala/wvlet/airframe/control/CircuitBreaker.scala
@@ -188,7 +188,7 @@ case class CircuitBreaker(
 
   /**
     * Set a fallback handler which process the exception happened in the code block.
-    * The default is just throw the exception as it is.
+    * The default is just throwing the exception as it is.
     */
   def withFallbackHandler(handler: Throwable => Any): CircuitBreaker = {
     this.copy(fallbackHandler = handler)
@@ -331,7 +331,7 @@ case class CircuitBreaker(
       case ResultClass.Succeeded(x) =>
         recordSuccess
         val clazz = implicitly[ClassTag[A]].runtimeClass
-        if (clazz.isInstance(x)) {
+        if (clazz.isAssignableFrom(x.getClass)) {
           x.asInstanceOf[A]
         } else {
           throw new ClassCastException(s"${x} is not an instance of ${resultClass}")
@@ -345,7 +345,7 @@ case class CircuitBreaker(
         }
         val x     = fallbackHandler(cause)
         val clazz = implicitly[ClassTag[A]].runtimeClass
-        if (clazz.isInstance(x)) {
+        if (clazz.isAssignableFrom(x.getClass)) {
           x.asInstanceOf[A]
         } else {
           throw new ClassCastException(s"${x} is not an instance of ${resultClass}")
--- a/airframe-control/src/main/scala/wvlet/airframe/control/CircuitBreaker.scala
+++ b/airframe-control/src/main/scala/wvlet/airframe/control/CircuitBreaker.scala
@@ -333,7 +333,7 @@ case class CircuitBreaker(
         result.get
       case ResultClass.Failed(retryable, cause, _) =>
         recordFailure(cause)
-        if(retryable) {
+        if (retryable) {
           // If the error is retryable, rethrow as it is then the caller (maybe Retryer) should handle it.
           throw cause
         } else {
--- a/airframe-control/src/main/scala/wvlet/airframe/control/CircuitBreaker.scala
+++ b/airframe-control/src/main/scala/wvlet/airframe/control/CircuitBreaker.scala
@@ -107,14 +107,14 @@ trait CircuitBreakerRecoveryPolicy {
 }
 
 object CircuitBreakerRecoveryPolicy {
-  def immediateRecoveryPolicy = new CircuitBreakerRecoveryPolicy() {
+  def recoverImmediately = new CircuitBreakerRecoveryPolicy() {
     override def recordSuccess: Unit = {}
     override def recordFailure: Unit = {}
     override def reset: Unit         = {}
     override def canRecover: Boolean = true
   }
 
-  def consequentSuccessRecoveryPolicy(numberOfSuccess: Int) = new CircuitBreakerRecoveryPolicy() {
+  def recoverAfterConsecutiveSuccesses(numberOfSuccess: Int) = new CircuitBreakerRecoveryPolicy() {
     private val counter              = new AtomicInteger(0)
     override def recordSuccess: Unit = counter.incrementAndGet()
     override def recordFailure: Unit = counter.set(0)
@@ -122,7 +122,7 @@ object CircuitBreakerRecoveryPolicy {
     override def canRecover: Boolean = counter.get() >= numberOfSuccess
   }
 
-  def elapsedTimeRecoveryPolicy(elapsedTimeMillis: Int) = new CircuitBreakerRecoveryPolicy() {
+  def recoverAfterWait(elapsedTimeMillis: Int) = new CircuitBreakerRecoveryPolicy() {
     private val timestamp            = new AtomicLong(Long.MaxValue)
     override def recordSuccess: Unit = timestamp.compareAndSet(Long.MaxValue, System.currentTimeMillis())
     override def recordFailure: Unit = timestamp.set(Long.MaxValue)
@@ -140,7 +140,7 @@ case class CircuitBreaker(
     onStateChangeListener: CircuitBreakerContext => Unit = CircuitBreaker.reportStateChange,
     fallbackHandler: Throwable => Any = t => throw t,
     delayAfterMarkedDead: RetryPolicy = new Jitter(new RetryPolicyConfig(initialIntervalMillis = 30000)), // 30 seconds
-    recoveryPolicy: CircuitBreakerRecoveryPolicy = CircuitBreakerRecoveryPolicy.immediateRecoveryPolicy,
+    recoveryPolicy: CircuitBreakerRecoveryPolicy = CircuitBreakerRecoveryPolicy.recoverImmediately,
     private var nextProvingTimeMillis: Long = Long.MaxValue,
     private var provingWaitTimeMillis: Long = 0L,
     var lastFailure: Option[Throwable] = None,
@@ -343,7 +343,9 @@ case class CircuitBreaker(
           if (clazz.isAssignableFrom(x.getClass)) {
             x.asInstanceOf[A]
           } else {
-            throw new ClassCastException(s"${x} is not an instance of ${resultClass}")
+            throw new ClassCastException(
+              s"The fallback handler is returning ${x}, which is not an instance of ${clazz.getName}"
+            )
           }
         }
     }
--- a/airframe-control/src/main/scala/wvlet/airframe/control/CircuitBreaker.scala
+++ b/airframe-control/src/main/scala/wvlet/airframe/control/CircuitBreaker.scala
@@ -13,16 +13,13 @@
  */
 package wvlet.airframe.control
 
-import java.util.concurrent.atomic.AtomicReference
-
-import wvlet.airframe.control.ResultClass.{Failed, Succeeded}
-import wvlet.airframe.control.Retry.RetryableFailure
+import java.util.concurrent.atomic.{AtomicInteger, AtomicLong, AtomicReference}
 
 import scala.util.{Failure, Success, Try}
 import wvlet.log.LogSupport
-import wvlet.airframe.control.Retry.RetryPolicy
-import wvlet.airframe.control.Retry.RetryPolicyConfig
-import wvlet.airframe.control.Retry.Jitter
+import wvlet.airframe.control.Retry.{Jitter, RetryPolicy, RetryPolicyConfig, RetryableFailure}
+
+import scala.reflect.ClassTag
 
 /**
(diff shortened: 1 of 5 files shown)


xerial:
@takezoe I tried the same approach in Retry (making it Retry[A]), but it didn't work well as it will restrict the result type unnecessarily.

xerial:
For example, requiring describing the input type of CircuitBreaker[A] or RetryContext[A] when returning retryableFailure[A] looks cumbersome as it's irrelevant to the error handler itself.

xerial:
@takezoe Previously I removed the requirement of having return type as type parameter from Retry to support Future type https://github.com/wvlet/airframe/pull/482

airframe-control/src/main/scala/wvlet/airframe/control/CircuitBreaker.scala
```diff
@@ -258,20 +260,22 @@ case class CircuitBreaker(
...
     */
-  def run[A](body: => A): Unit = {
+  def run[A](body: => A): A = {
     verifyConnection
 
     val result = Try(body)
     val resultClass = result match {
       case Success(x) => resultClassifier(x)
       case Failure(RetryableFailure(e)) =>
+        // TODO Returning retryableFailure seems meaningless in the current implementation. What's your intention of this?
```
xerial: We should add ResultClass.ignorableFailure(e) instead

takezoe:
I gave up to generalize `ResultClass` as `ResultClass[A]`. Just added a result value to `ResultClass.Succeeded` as `Any`.

airframe-control/src/main/scala/wvlet/airframe/control/Retry.scala
```diff
@@ -279,9 +280,14 @@ object Retry extends LogSupport {
         }
 
         resultClass match {
-          case ResultClass.Succeeded =>
+          case ResultClass.Succeeded(x) =>
             // OK. Exit the loop
-            result = Some(ret.get)
+            val clazz = implicitly[ClassTag[A]].runtimeClass
+            if (clazz.isInstance(x)) {
```
xerial: Using isInstance is not safe if the result is a subclass of A:
https://stackoverflow.com/questions/3949260/java-class-isinstance-vs-class-isassignablefrom

Use isAssingableFrom

airframe-control/src/main/scala/wvlet/airframe/control/CircuitBreaker.scala
```diff
@@ -140,7 +187,15 @@ case class CircuitBreaker(
   }
 
   /**
-    * Set an event listner that monitors CircuitBreaker state changes
+    * Set a fallback handler which process the exception happened in the code block.
+    * The default is just throw the exception as it is.
```
xerial: s/throw/throwing/

airframe-control/src/main/scala/wvlet/airframe/control/ResultClass.scala
```diff
@@ -25,7 +25,7 @@ object ResultClass {
   /**
     * A label for the successful code execution result
     */
-  case object Succeeded extends ResultClass
+  case class Succeeded(value: Any) extends ResultClass
```
xerial: This will extend the responsibility of the result classifier to generate the default result to convert Failed(exception) into Succeeded(value), so if we add this, the class will no longer ResultClass. 

Ideally, we should minimize the responsibility of the result classifier only to _classify_ result types, and let the failback handler produce the default response for certain types of errors.

airframe-control/src/main/scala/wvlet/airframe/control/CircuitBreaker.scala
```diff
@@ -248,34 +306,50 @@ case class CircuitBreaker(
...
+          x.asInstanceOf[A]
+        } else {
+          throw new ClassCastException(s"${x} is not an instance of ${resultClass}")
+        }
+      case ResultClass.Failed(retryable, cause, _) =>
+        if (!retryable) {
+          recordFailure(cause)
+        } else {
+          // TODO should record as success in this case?
+          recordSuccess
```
xerial: I don't think so. We should record failures for both cases. For example, network connection failure is a retryable error but if it continues, we should cut the circuit.

airframe-control/src/main/scala/wvlet/airframe/control/CircuitBreaker.scala
```diff
@@ -248,34 +306,50 @@ case class CircuitBreaker(
...
-      case Failure(e) =>
-        errorClassifier(e)
+      case Success(x)                   => resultClassifier(x)
+      case Failure(RetryableFailure(e)) => ResultClass.retryableFailure(e)
+      case Failure(e)                   => errorClassifier(e)
     }
+
     resultClass match {
-      case Succeeded =>
+      case ResultClass.Succeeded(x) =>
```
xerial: I'm against embedding the result object here as we should have it in result variable. Why can't we use only failbackHandler?

airframe-control/src/main/scala/wvlet/airframe/control/CircuitBreaker.scala
takezoe: If so, what's the meaning of ignoreable (or retryable)?

airframe-control/src/main/scala/wvlet/airframe/control/CircuitBreaker.scala
takezoe: This is because of errorClassifier can convert `Failure` to `Succeeded` and it causes `RuntimeError` at https://github.com/wvlet/airframe/blob/dafedf03e500ad140cfab28f3dd76388caa8f6bb/airframe-control/src/main/scala/wvlet/airframe/control/CircuitBreaker.scala#L275

I once thought that modifying errorClassifier to be allowed only classifying retryable or not here, but the same thing happens in `Retry`.
https://github.com/wvlet/airframe/blob/dafedf03e500ad140cfab28f3dd76388caa8f6bb/airframe-control/src/main/scala/wvlet/airframe/control/Retry.scala#L284

I would say the current structure is broken. We need to change something to fix anyway but I'm not sure what's the best solution covering both `CircuitBreaker` and `Retry`.

> Why can't we use only failbackHandler?

How fallbackHandler can help this situation?

takezoe: Maybe errorClassifier should be `Throwable => ResultClass.Failed`?

airframe-control/src/main/scala/wvlet/airframe/control/CircuitBreaker.scala
takezoe: Fixed.

airframe-control/src/main/scala/wvlet/airframe/control/Retry.scala
takezoe: Fixed.

takezoe:
Hmm... Tests are blocked in GitHub Actions in this pull request. Need investigation.

airframe-control/src/main/scala/wvlet/airframe/control/CircuitBreaker.scala
xerial: What I think is:
• ResultClassifier should not generate any result, so converting Failure to Succeeded should never happen. 
• When we see Failed(retryable/ignorable = true), we just throw this exception so that Retry can submit the request again.
• When we see Failed(retryable = false), we should apply a fallback handler to produce some fallback results. The default behavior is just throwing an exception. If the user provides a custom default result, we should return that by casting it to A.

airframe-control/src/main/scala/wvlet/airframe/control/CircuitBreaker.scala
xerial: Ignorable is just one of the failure types, which doesn't need to report the exception somewhere else. It's not for controlling the behavior of CircuitBreaker. 

Some use case I found in Finagle is if a back-up request succeeds, the exception happened in the original request can be ignored.

airframe-control/src/main/scala/wvlet/airframe/control/CircuitBreaker.scala
takezoe: > ResultClassifier should not generate any result, so converting Failure to Succeeded should never happen. Try.Failure can be classified only for Failed(retryable = true or false)

Just confirmation. We have two types of classifiers `resultClassifier` and `errorClassifier`. I'd like to clarify the role of them.

- `errorClassifier` is responsible for determining if error is ignoreable(= retryable). He can not convert failure as success.
- On the other hand, still `resultClassifier` can convert success as failure?

> Question: If the user really needs to classify some Failure response as Succeeded, should we apply the fallback handler to produce the default result?

Not sure, but isolating the classifying result logic and the producing default value logic might make the situation hard if we need to produce the default value depending on the error type.

takezoe:
@xerial Could you take a look again?

airframe-control/src/main/scala/wvlet/airframe/control/CircuitBreaker.scala
```diff
@@ -86,14 +83,64 @@ trait CircuitBreakerContext {
...
+  def reset: Unit
+
+  /**
+    * Check if CircuitBreaker can recover from HALF_OPEN to CLOSE.
+    */
+  def canRecover: Boolean
+}
+
+object CircuitBreakerRecoveryPolicy {
+  def immediateRecoveryPolicy = new CircuitBreakerRecoveryPolicy() {
```
xerial: `recoverImmediately`

airframe-control/src/main/scala/wvlet/airframe/control/CircuitBreaker.scala
```diff
@@ -86,14 +83,64 @@ trait CircuitBreakerContext {
...
+    override def recordSuccess: Unit = {}
+    override def recordFailure: Unit = {}
+    override def reset: Unit         = {}
+    override def canRecover: Boolean = true
+  }
+
+  def consequentSuccessRecoveryPolicy(numberOfSuccess: Int) = new CircuitBreakerRecoveryPolicy() {
```
xerial: `recoverAfterConsecutiveSuccesses`

airframe-control/src/main/scala/wvlet/airframe/control/CircuitBreaker.scala
```diff
@@ -86,14 +83,64 @@ trait CircuitBreakerContext {
...
+    private val counter              = new AtomicInteger(0)
+    override def recordSuccess: Unit = counter.incrementAndGet()
+    override def recordFailure: Unit = counter.set(0)
+    override def reset: Unit         = counter.set(0)
+    override def canRecover: Boolean = counter.get() >= numberOfSuccess
+  }
+
+  def elapsedTimeRecoveryPolicy(elapsedTimeMillis: Int) = new CircuitBreakerRecoveryPolicy() {
```
xerial: `recoverAfterWait`

airframe-control/src/main/scala/wvlet/airframe/control/CircuitBreaker.scala
```diff
@@ -248,34 +306,46 @@ case class CircuitBreaker(
...
+          // If the error is retryable, rethrow as it is then the caller (maybe Retryer) should handle it.
+          throw cause
+        } else {
+          // If the error is not retryable, apply fallbackHandler
+          val x     = fallbackHandler(cause)
+          val clazz = implicitly[ClassTag[A]].runtimeClass
+          if (clazz.isAssignableFrom(x.getClass)) {
+            x.asInstanceOf[A]
+          } else {
+            throw new ClassCastException(s"${x} is not an instance of ${resultClass}")
```
xerial: The error message needs to show how to fix it:
`Fallback handler is returning ${x}, which is not an instance of ...`

xerial:
OK. Added naming suggestions. Overall it LGTM

airframe-control/src/main/scala/wvlet/airframe/control/ResultClass.scala
takezoe: I see. Modified errorClassifier can return only `ResultClass.Failed` so that there is no necessity to have a result value at ResultClass.

airframe-control/src/main/scala/wvlet/airframe/control/CircuitBreaker.scala
takezoe: Got it. Thanks!

airframe-control/src/main/scala/wvlet/airframe/control/CircuitBreaker.scala
takezoe: Fixed.

airframe-control/src/main/scala/wvlet/airframe/control/CircuitBreaker.scala
takezoe: Fixed.

airframe-control/src/main/scala/wvlet/airframe/control/CircuitBreaker.scala
takezoe: Fixed.

airframe-control/src/main/scala/wvlet/airframe/control/CircuitBreaker.scala
takezoe: Fixed.

takezoe:
Thanks! I adopted all your suggestions. Will merge after CI passed.

Issue #901, opened 2020-01-19
Design.showConfig throws java.lang.NoSuchFieldException: port


takezoe:
```
[info]   java.lang.NoSuchFieldException: port
[info]   at java.lang.Class.getDeclaredField(Class.java:2070)
[info]   at wvlet.airframe.surface.reflect.RuntimeMethodParameter.field$lzycompute(RuntimeMethodParameter.scala:33)
[info]   at wvlet.airframe.surface.reflect.RuntimeMethodParameter.field(RuntimeMethodParameter.scala:33)
[info]   at wvlet.airframe.surface.reflect.RuntimeMethodParameter.get(RuntimeMethodParameter.scala:37)
[info]   at wvlet.airframe.config.Config.$anonfun$printConfig$3(Config.scala:102)
[info]   at scala.collection.Iterator.foreach(Iterator.scala:941)
[info]   at scala.collection.Iterator.foreach$(Iterator.scala:941)
[info]   at scala.collection.AbstractIterator.foreach(Iterator.scala:1429)
[info]   at scala.collection.IterableLike.foreach(IterableLike.scala:74)
[info]   at scala.collection.IterableLike.foreach$(IterableLike.scala:73)
[info]   ...
```
Looks like the root cause is https://github.com/wvlet/airframe/pull/858 since this change includes making `DBConfig.port` private.

takezoe:
@xerial Added a test case which can reproduce this issue at: https://github.com/wvlet/airframe/pull/902 and wrote possible options there. Could you take a look?

xerial:
FYI

```
$ javap -p airframe-config/target/scala-2.12/test-classes/wvlet/airframe/config/AirframeBootstrapTest\$DBConfig.class
Compiled from "AirframeBootstrapTest.scala"
public class wvlet.airframe.config.AirframeBootstrapTest$DBConfig implements scala.Product,scala.Serializable {
  private final java.lang.String host;
  private final scala.Option<java.lang.Object> wvlet$airframe$config$AirframeBootstrapTest$DBConfig$$port;  <<--- special field name is used if the default value is defined
  public scala.Option<java.lang.Object> port$access$1();
  public java.lang.String host();
  public scala.Option<java.lang.Object> wvlet$airframe$config$AirframeBootstrapTest$DBConfig$$port();
  public wvlet.airframe.config.AirframeBootstrapTest$DBConfig copy(java.lang.String, scala.Option<java.lang.Object>);
[...]

PR #902, opened 2020-01-20
Add test case for showConfig()

takezoe:
This test case reproduces https://github.com/wvlet/airframe/issues/901

Possible options are:
- Config must not have `private val` in the constructor
- Exclude `private val` fields from the target of `showConfig()`


xerial:
Closing this in favor of #907

PR #943, opened 2020-02-16
Add sbt-airframe plugin

xerial:
- Support generating HTTP client code from airframe-http interfaces
- [x] plugin settings
- [x] Read http interfaces 
- [x] Generate http client code for JVM
- [x] Generate Scala.js client code


xerial:
To support Scala.js code generation, we should make airframe-control / airframe-http cross projects for Scala JVM/Scala.js. I'll work on it in another PR.

PR #946, opened 2020-02-18
airframe-http-okhttp client

shimamoto:
Add an HTTP client that uses OkHttp for synchronous calls.


shimamoto:
@xerial @takezoe Could you take a look when you have time?

xerial:
ok. This code basically LGTM. I need some modifications for using it with sbt-airframe. I'll merge this PR after that.

PR #950, opened 2020-02-26
[DO NOT MERGE] Add Node.js facade

xerial:


xerial:
Closes as we have no plan on working on this.

PR #1008, opened 2020-03-24
airframe-okhttp: Fix for Scala 2.11 and release process in build.sbt

xerial:
I tried to upgrade Airframe version in td-scala-util and found airframe-http-okhttp was not available in Maven. So I tried to publish this module individually, and found this Scala 2.11 incompatibility issue.

### DIFF
--- a/airframe-http-okhttp/src/main/scala/wvlet/airframe/http/okhttp/OkHttpRetryInterceptor.scala
+++ b/airframe-http-okhttp/src/main/scala/wvlet/airframe/http/okhttp/OkHttpRetryInterceptor.scala
@@ -5,7 +5,8 @@ import wvlet.airframe.control.ResultClass
 import wvlet.airframe.control.Retry.RetryContext
 import wvlet.airframe.http.HttpClientMaxRetryException
 
-import scala.util.Try
+import scala.util.control.NonFatal
+import scala.util.{Failure, Success, Try}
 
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -152,6 +152,7 @@ lazy val jvmProjects: Seq[ProjectReference] = communityBuildProjects ++ Seq[Proj
   fluentd,
   airspecLight,
   finagle,
+  okhttp,
```
xerial: airframe-okhttp was not included in airframe 20.3.2 release. I'll release 20.3.3 after merging this PR

airframe-http-okhttp/src/main/scala/wvlet/airframe/http/okhttp/OkHttpRetryInterceptor.scala
```diff
@@ -5,19 +5,28 @@ import wvlet.airframe.control.ResultClass
...
+          retryContext.resultClassifier(r)
+        } catch {
+          case NonFatal(e) =>
+            retryContext.errorClassifier(e)
+        }
+      case Failure(e) =>
+        retryContext.errorClassifier(e)
+    }
 
-    response.fold(
```
xerial: Unfortunately, `fold` was not available in Scala 2.11

shimamoto: I wasn't aware of that. thank you

PR #1017, opened 2020-03-28
airframe-http: Simplify Request/Response adapters

xerial:
- Remove depreacated SimpleHttpRequest/Response
- Use HttpMessage.Message for managing content body
- Add headerOf(x) to adapters
- This change will allow using HttpMessage.Request in RPC interface argumetns, and the interface can be used for both Scala.js (client) and Scala JVM (server) implementations


xerial:
cc: @shimamoto

PR #1023, opened 2020-04-01
Break execution in Parallel

takezoe:
Allow to break execution in Parallel as follows:
```scala
val result = Parallel.run(Seq(1, 2, 3), parallelism = 1) { i =>
  if (i == 2) {
    Parallel.break
  }
  i
}
```


xerial:
ok. Merged

Issue #1025, opened 2020-04-03
airframe-http: Custom error responses


xerial:
When handling errors, we may need to return custom error responses:

```scala
Http.response(status).withJsonContent(json)  // return json body
Http.serverException(request, NotFound_404).withJson(json) // return json response
Http.serverException(request, NotFound_404).withContent[A](a)  // switch json or msgpack by looking at Accept header

def rpcMethod(request:Request): A = {
   try {
      new A(...) // normal path
   }
   catch {
     // Error path
     case e:Exception => throw Http.serverException(request, status).withContent[ErrorMessageObject)(err) 
   }
} 

```

xerial:
We also need `throw Http.redirect(location url)` to redirect requests.

xerial:
Added in #1036

PR #1036, opened 2020-04-14
airframe-http: #1025 Support custom server exception response

xerial:
- Add Http.serverException(...) methods so that we can customize error responses at ease
- Support built-in conversion from error response objects into JSON/MessagePack
- Add a handy method to implement 302 redirect (cc: @yuokada)


airframe-http/src/main/scala/wvlet/airframe/http/Http.scala
```diff
@@ -37,6 +39,51 @@ object Http {
...
   }
+
+  /**
+    * Create an exception to redirect (status code = 302) the request to the target locationUrl
+    * @param locationUrl
+    * @param status
+    * @return
+    */
+  def redirectException(
+      locationUrl: String,
```
yuokada, line 50: This method supports the relative redirect 👍

PR #1052, opened 2020-05-04
Use sbt-scalafmt for code format tests

xerial:
Upgrading scalfmt binary for each scalafmt version upgrade is a bit tedious task. Use sbt-scalafmt instead.


xerial:
@Pritesh-Patel Thanks for the first review of Airframe :)

PR #1059, opened 2020-05-07
Fix the document about airframe-jdbc

amaya382:
Use `newConnectionPool` instead of `apply`.

https://github.com/wvlet/airframe/blob/d2e9e05e8caf0bb07f7ef1120b6846691577b44e/airframe-jdbc/src/main/scala/wvlet/airframe/jdbc/ConnectionPoolFactory.scala#L42-L44


xerial:
Thanks for the fix!

PR #1060, opened 2020-05-07
airframe-http: Handle when port number is empty string

xerial:


xerial:
@shimamoto Found this bug when using this JS client on the server behind ALB

xerial:
@shimamoto I think you already implemented a similar workaround in our internal code.

PR #1085, opened 2020-05-15
airframe-http: Fix query_string parsing of HttpMessage.Request

xerial:


airframe-http-finagle/src/main/scala/wvlet/airframe/http/finagle/package.scala
```diff
@@ -82,24 +82,10 @@ package object finagle {
...
-    override def headerOf(request: Request): HttpMultiMap = {
-      val h = request.headerMap
-      var m = HttpMultiMap.empty
-      for (k <- h.keys) {
-        h.getAll(k).map { v => m = m.add(k, v) }
-      }
-      m
-    }
-    override def queryOf(request: Request): HttpMultiMap = {
-      val p = request.params
```
xerial, line 96: This is an additional fix as Finagle's Requet.params will look at content body too. We need to extract query parameters from the uri of the request.

PR #1087, opened 2020-05-15
airframe-http: #1086 Support query_string mapping for non GET requests

### LINKED ISSUE #1086: airframe-http: query_string mapping support for non-GET requests
Mapping query_string parameters to function arguments is supported only for GET requests. If we want to pass extra parameters (e.g., session_id), mapping them to function args will be useful.

```scala
@Endpoint(method = HttpMethod.POST, path="/v1/user/:id")
def add(id:String, session_id: Option[String]) = {
   ... 
}
```

With this support `POST /v1/user/1?session_id=xxxx` will call `add(1, Some(xxx))`.

xerial:
This will close #1086


xerial:
cc: @takezoe @shimamoto

PR #1089, opened 2020-05-16
airframe-http: Retry on Finagle's ChannelClosedException

xerial:
When a server is not yet ready, FinagleClient will throw ChannelClosedException. This happens, for example, when launching Nginx inside a test suite, and trying to connect it using FinagleClient before Nginx is not yet started inside the container. We can wait until a docker container starts, but waiting the service startup inside the container needs extra wait. 

By adding this retry pattern, we can retry http requests to a docker container until the application inside the container actually starts.

### DIFF
--- a/airframe-http/src/main/scala/wvlet/airframe/http/HttpClientException.scala
+++ b/airframe-http/src/main/scala/wvlet/airframe/http/HttpClientException.scala
@@ -15,6 +15,7 @@ package wvlet.airframe.http
 import java.io.{EOFException, IOException}
 import java.lang.reflect.InvocationTargetException
 import java.net._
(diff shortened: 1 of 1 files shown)


airframe-http/src/main/scala/wvlet/airframe/http/HttpClientException.scala
```diff
@@ -177,8 +182,17 @@ object HttpClientException extends LogSupport {
         case other =>
           nonRetryableFailure(e)
       }
+    // Exceptions from Finagle. Using the string class names so as not to include Finagle dependencies.
```
takezoe, line 185: 👍

PR #1098, opened 2020-05-18
airframe-codec: #1091 Support complex types in AnyCodec

### LINKED ISSUE #1091: airframe-codec: Support more complex types in AnyCodec
The current codec uses toString for complex types:
```scala
case class User(id:Int, name:String)

// MessageCodec.of[Map[String, Any]].toJson
Map("user-> User(1, "leo")) ->. {"user":"User(1,\"leo\")"}
```

In Scala, it is possible to find a codec using the runtime reflection, so AnyCodec can produce precise output:
```scala
case x:Any => 
  val cls = x.getClass
  val codec = ReflectSurface.ofClass(cls)
  codec.toMsgPack(cls.cast(x)) 

Map("user-> User(1, "leo")) ->. {"user":{"id":1, "name":"leo"}}
```

xerial:
This will close #1091

@shimamoto @takezoe This PR will be useful to report the detailed RPC call arguments in JSON format.

### DIFF
--- a/airframe-codec/.jvm/src/main/scala/wvlet/airframe/codec/Compat.scala
+++ b/airframe-codec/.jvm/src/main/scala/wvlet/airframe/codec/Compat.scala
@@ -52,9 +52,14 @@ object Compat {
       cl: Class[_],
       codecFactory: MessageCodecFactory = MessageCodecFactory.defaultFactoryForJSON
   ): Option[MessageCodec[_]] = {
+    // Finding the surface using reflection
     ReflectSurfaceFactory.ofClass(cl) match {
-      case g: GenericSurface if g.params == 0 => None
-      case surface                            => Some(codecFactory.of(surface))
+      case g: GenericSurface if g.params == 0 =>
+        // If this type has no parameters, do not use ObjectCodec for JSON output
+        None
+      case surface =>
+        // Otherwise, use the regular codec finder from Surface
+        Some(codecFactory.of(surface))
     }
   }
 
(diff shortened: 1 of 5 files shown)


airframe-codec/.jvm/src/main/scala/wvlet/airframe/codec/Compat.scala
```diff
@@ -48,6 +48,20 @@ object Compat {
     JavaTimeCodec.javaTimeCodecs ++ JavaStandardCodec.javaStandardCodecs
 
   def codecOf[A: ru.TypeTag]: MessageCodec[A] = MessageCodecFactory.defaultFactory.of[A]
+  def codecOfClass(
+      cl: Class[_],
+      codecFactory: MessageCodecFactory = MessageCodecFactory.defaultFactoryForJSON
+  ): Option[MessageCodec[_]] = {
+    // Finding the surface using reflection
+    ReflectSurfaceFactory.ofClass(cl) match {
+      case g: GenericSurface if g.params == 0 =>
```
shimamoto, line 57: I think this can work in most cases, it's okay. Just a thought, for instance, case object is not covered?

xerial: case objects will be resolved as EnumSurface so this pass will not be used, but let me double-check that.

xerial: And that was a good catch. It didn't work for case objects. I revised this logic to be specific to the case when ObjectCodec or ObjectMapCodec has empty parameters.

xerial:
Added a test case for case objects.

PR #1099, opened 2020-05-18
airframe-surface, codec: #1092 Support Enum-like classes in Scala.js

### LINKED ISSUE #1092: airframe-surface: Surface[X].getStringExtractor: Option[String => Option[X]]
In Scala.js, we cannot find X.unapply(String) method using reflection, so enum-like case objects cannot be serialized/deserialized without using custom codecs. 

unapply method can be found in compile time, so we should provide it in Surface.

```scala
sealed trait Suit
case object Spade extends Suit
case object Heart extends Suit

object Suit {
  def values: Seq[Suit] = Seq(Spade, Heart)
  def unapply(s:String): Option[Suit] = values.find(_.toString == s)
}

MessageCodec.of[Suit] // Use Surface.of[Suit].getStringExtracter for String input values.
```

xerial:
Support enum like classes with Surface and codec. This PR closes #1092

This support will be useful for defining RPC/Endpoint model classes with Enum-like case objects that can be shared between Scala JVM and Scala.js. 

cc: @takezoe @shimamoto

### DIFF
--- a/airframe-surface/jvm/src/main/scala/wvlet/airframe/surface/reflect/ReflectSurfaceFactory.scala
+++ b/airframe-surface/jvm/src/main/scala/wvlet/airframe/surface/reflect/ReflectSurfaceFactory.scala
@@ -687,6 +687,4 @@ object ReflectSurfaceFactory extends LogSupport {
       }
     }
   }
-
-  class ReflectEnumSurface(override val rawType: Class[_]) extends GenericSurface(rawType) {}
 }
--- a/airframe-surface/jvm/src/main/scala/wvlet/airframe/surface/reflect/ReflectSurfaceFactory.scala
+++ b/airframe-surface/jvm/src/main/scala/wvlet/airframe/surface/reflect/ReflectSurfaceFactory.scala
@@ -19,7 +19,6 @@ import java.util.concurrent.ConcurrentHashMap
 import wvlet.airframe.surface._
 import wvlet.log.LogSupport
 
-import scala.collection.mutable
 import scala.jdk.CollectionConverters._
 import scala.reflect.runtime.{universe => ru}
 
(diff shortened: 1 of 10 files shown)


airframe-surface/shared/src/main/scala/wvlet/airframe/surface/SurfaceMacros.scala
```diff
@@ -571,6 +590,21 @@ private[surface] object SurfaceMacros {
...
         newGenericSurfaceOf(baseType)
+      case t if hasStringUnapply(t) =>
+        t.typeSymbol.fullName.split("\\.").toList match {
+          case p1 :: tl =>
+            // companion object package -> Select(Select( .... ))) tree
+            val ex = tl.foldLeft[Tree](Ident(TermName(p1))) { (z, n) =>
+              Select(z, TermName(n))
+            }
+            val expr = q"""wvlet.airframe.surface.EnumSurface(classOf[${finalTypeOf(t)}], { (cl: Class[_], s: String) =>
+                ${ex}.unapply(s).asInstanceOf[Option[Any]]
```
shimamoto, line 601: We cannot find unapply method at runtime, which is why Tree is used instead of TypeConverter.convert?

airframe-surface/shared/src/main/scala/wvlet/airframe/surface/SurfaceMacros.scala
```diff
@@ -571,6 +590,21 @@ private[surface] object SurfaceMacros {
...
+              }
+            )"""
+            expr
+          case _ =>
+            newGenericSurfaceOf(t)
```
shimamoto, line 606: This means we cannot define Enum-like classes in the default package? But in reality, we almost always define a package.

shimamoto:
To use this support, the requirements are

- A companion object's unapply method
- toString should return a string that can be extracted by an unapply method

Is this correct?

airframe-surface/jvm/src/main/scala/wvlet/airframe/surface/reflect/ReflectSurfaceFactory.scala
```diff
@@ -671,4 +687,6 @@ object ReflectSurfaceFactory extends LogSupport {
       }
     }
   }
+
+  class ReflectEnumSurface(override val rawType: Class[_]) extends GenericSurface(rawType) {}
```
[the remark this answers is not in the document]
xerial: Thanks. This is a garbage class. I'll remove it.

airframe-surface/shared/src/main/scala/wvlet/airframe/surface/SurfaceMacros.scala
xerial: Yes.

airframe-surface/shared/src/main/scala/wvlet/airframe/surface/SurfaceMacros.scala
xerial: OK. I'll add support for unapply methods in package objects.

xerial:
@shimamoto Yes. These are correct. I'll try adding (package).unapply(s: String) too

airframe-surface/shared/src/main/scala/wvlet/airframe/surface/SurfaceMacros.scala
```diff
@@ -571,6 +590,21 @@ private[surface] object SurfaceMacros {
         q"wvlet.airframe.surface.ExistentialType"
       case t @ RefinedType(List(_, baseType), decl) =>
         newGenericSurfaceOf(baseType)
+      case t if hasStringUnapply(t) =>
+        t.typeSymbol.fullName.split("\\.").toList match {
+          case p1 :: tl =>
+            // companion object package -> Select(Select( .... ))) tree
+            val ex = tl.foldLeft[Tree](Ident(TermName(p1))) { (z, n) =>
+              Select(z, TermName(n))
+            }
```
takezoe, line 599: This generates a tree like `Select(Select(Ident(TermName(java)), TermName(obj)), TermName(Object))` for `java.lang.Object` and it selects a companion object. I see...

xerial: There might be a better way to produce this expression, but I could not find it...

xerial:
@shimamoto I've tried to find package object in Scala Macros, but it was difficult for me to do that. I'm not sure how frequent it is to define `unapply(s:String): Option[X]` in package object, though. And also, package object will be dropped in Dotty (Scala 3), so we might not need to support this use case https://dotty.epfl.ch/docs/reference/dropped-features/package-objects.html

PR #1114, opened 2020-05-31
airframe-http: #1112 Report RPC interface class

### LINKED ISSUE #1112: airframe-http: http_access log should report the base RPC class (rpc_interface)
http_access.json reports rpc_class values using its implementation class names. However, the reporting the root class that defines`@RPC` annotation is preferred for consistency of logging regardless of RPC or Endpoint implementation names. So reporting both rpc_interface and rpc_class is better.

xerial:
Closes #1112

@shimamoto This will be useful for collecting audit logs of RPC calls.

### DIFF
--- a/airframe-http/.jvm/src/main/scala/wvlet/airframe/http/Router.scala
+++ b/airframe-http/.jvm/src/main/scala/wvlet/airframe/http/Router.scala
@@ -140,7 +140,15 @@ case class Router(
             .map { m => (m, m.findAnnotationOf[Endpoint]) }
             .collect {
               case (m: ReflectMethodSurface, Some(endPoint)) =>
-                ControllerRoute(controllerSurface, endPoint.method(), prefixPath + endPoint.path(), m)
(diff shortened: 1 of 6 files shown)


airframe-http/.jvm/src/main/scala/wvlet/airframe/http/Router.scala
```diff
@@ -140,7 +140,15 @@ case class Router(
...
-                ControllerRoute(controllerSurface, endPoint.method(), prefixPath + endPoint.path(), m)
+                val endpointInterfaceCls =
+                  controllerSurface.findAnnotationOwnerOf[Endpoint].getOrElse(controllerSurface.rawType)
+                ControllerRoute(
+                  endpointInterfaceCls,
+                  controllerSurface,
+                  endPoint.method(),
+                  prefixPath + endPoint.path(),
+                  m
+                )
```
shimamoto, line 151: Ah, this is for conventional `@Endpoint`. I see

Issue #1104, opened 2020-05-23
airframe-metric: succinctDuration(Double)


xerial:

xerial:
We can use Double.round instead

PR #1116, opened 2020-05-31
airframe-metrics: #1104 Add ElapsedTime.succinctMillis(Double)

xerial:


xerial:
It seems we don't need to have this. Just using Double.round seems enough

Issue #1100, opened 2020-05-18
airframe-http: RPC binds raw JSON request body to a String type function argument


xerial:
## Problem description
```scala
@RPC 
trait Service {
  def hello(name:String): Response
}

client.service.hello("world") -> creates {"name":"world"} request body

hello({"name":"world"}) will be called
```

## Details

The current workaround is wrapping the RPC argument with a model class. However, this is not intuitive. So we should be able to extract the parameter list from the JSON body at this part: https://github.com/wvlet/airframe/blob/3a2d703c18858b339bf1ad92b8a7e867aaa486af/airframe-http/.jvm/src/main/scala/wvlet/airframe/http/router/HttpRequestMapper.scala#L82-L103
This code implicitly assumes that the content body will be mapped to a single argument for non GET requests. 

POST requests usually do not depend on query string parameters and we expect POST to send request object using the content body. So sbt-airframe generates JSON request body for primitive values because we need to wrap them with a JSON object (or array) for application/json  content type: https://github.com/wvlet/airframe/blob/3a2d703c18858b339bf1ad92b8a7e867aaa486af/airframe-http/.jvm/src/main/scala/wvlet/airframe/http/codegen/HttpClientIR.scala#L145-L149

If the RPC interface has primitive type values, using query string can be an option now that we have a fix for #1087. Or we should assume the content body has Map[String, Any] type like ` {"name":"world"} `.

## Possible solution

We need to clarify how to map HTTP requests to RPC/Endpoint calls in the protocol:

### Unary functions (1 argument, except Request and HttpContect)

`def method(p:T): R`

The current protocol is as follows:
- Rule 1) If T is a primitive type, read it from the query_string (for GET) or the request body (for POST) using JSON/MsgPack representation Map("p" -> value). This difference comes from the requirement that GET requests should not convey request bodies in general.
- Rule 2) If T is a complex type (not a primitive), read its parameters of T from the query_string (for GET) or read the entire T from the request body (for POST) described in JSON/MsgPack.

It seems Rule 2 is causing some inconsistency, so we may need to consolidate these two rules by requiring HTTP client to wrap body contents with JSON object (or MessgePack Map):

- New Rule:  RPC/Endpoint requests should describe the method arguments in Map(parameter_name -> JSON/msgpack) format in query_string (for GET) or query_string + request body (for POST/PUT/DELETE/PATCH). 
[...]

Issue #978, opened 2020-03-12
airframe-http: Validation for unnecessary attributes


xerial:
model class
```scala
case class OpsRequest(@required command:String, repeat:Boolean = false)
```

```json
{ 
  "command":"run", 
  "repet": true <== mistype of the input parameter
}
```

```scala
// This request will be passed to the server by ignoring "repet" input parameter:
OpsRequest(command = "run", repeat= false) 
```

In general, it should return 400 to avoid such a programming error.

PR #1126, opened 2020-06-10
airframe-http: #1100 Support unary and n-ary functions in RPC 

xerial:
Addresses #1100

- [x] Support an arbitrary number of function arguments for RPC calls.
- [x] sbt-airframe: If RPC call requires multiple arguments, wrap them into a single case class and send it as JSON or MsgPack. 
- [x] Changed to throw IllegalArgumentException and 400 (BadRequest) response when incompatible type data is used in HTTP requests. Note this change still doesn't cover cases described in #978 (unnecessary attributes are given) 
- [x] Properly handle application/octet-stream requests so as not to bind uploaded file contents to RPC arguments

### DIFF
--- a/airframe-http/.jvm/src/main/scala/wvlet/airframe/http/router/HttpRequestMapper.scala
+++ b/airframe-http/.jvm/src/main/scala/wvlet/airframe/http/router/HttpRequestMapper.scala
@@ -194,10 +194,12 @@ object HttpRequestMapper extends LogSupport {
               val mapValue = toCanonicalKeyNameMap(m)
               while (remainingArgs.nonEmpty) {
                 val arg = remainingArgs.head
-                val argValueOpt: Option[Any] = mapValue.get(CName.toCanonicalName(arg.name)).flatMap { x =>
-                  val argCodec = codecFactory.of(arg.surface)
-                  Some(argCodec.fromMsgPack(x.toMsgpack))
-                }
+                val argValueOpt: Option[Any] = mapValue
+                  .get(CName.toCanonicalName(arg.name))
+                  .map { x =>
+                    val argCodec = codecFactory.of(arg.surface)
+                    argCodec.fromMsgPack(x.toMsgpack)
+                  }
                 setValue(arg, argValueOpt)
                 remainingArgs = remainingArgs.tail
               }
--- a/airframe-http/.jvm/src/main/scala/wvlet/airframe/http/router/HttpRequestMapper.scala
+++ b/airframe-http/.jvm/src/main/scala/wvlet/airframe/http/router/HttpRequestMapper.scala
@@ -108,7 +108,7 @@ object HttpRequestMapper extends LogSupport {
 
         // Build the method argument instance from the query strings for GET requests
         argSurface match {
-          // For primitive type, Seq[Primitive], and Option[Primitive] values,
+          // For primitive type, Seq[Primitive], and Option[Primitive], Option[Seq[Primitive]] values,
           // it should already be found in the query strings, so bind None here
           case _ if argSurface.isPrimitive || isPrimitiveSeq(argSurface) =>
             setValue(arg, None)
--- a/airframe-http/.jvm/src/main/scala/wvlet/airframe/http/router/HttpRequestMapper.scala
+++ b/airframe-http/.jvm/src/main/scala/wvlet/airframe/http/router/HttpRequestMapper.scala
@@ -14,11 +14,10 @@
 package wvlet.airframe.http.router
 
 import wvlet.airframe.codec.PrimitiveCodec.StringCodec
-import wvlet.airframe.codec.{JSONCodec, MessageCodecFactory, ObjectCodec, ParamListCodec}
-import wvlet.airframe.http.HttpHeader.MediaType
+import wvlet.airframe.codec.{JSONCodec, MessageCodecFactory}
 import wvlet.airframe.http._
 import wvlet.airframe.json.JSON
-import wvlet.airframe.msgpack.spi.Value.{MapValue, StringValue}
+import wvlet.airframe.msgpack.spi.Value.MapValue
 import wvlet.airframe.msgpack.spi.{MessagePack, MsgPack, ValueFactory}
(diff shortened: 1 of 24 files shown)


xerial:
@takezoe @shimamoto I revised airframe-http Endpoint/RPC mapping so that we can support any type of functions including unary, n-ary functions even though primitive type values are involved.

I know this PR is fairly complicated, but this change will affect all airframe-http applications, so it's better to have your look.

airframe-http/.jvm/src/main/scala/wvlet/airframe/http/router/HttpRequestMapper.scala
```diff
@@ -40,84 +47,186 @@ object HttpRequestMapper extends LogSupport {
...
+              case m: MapValue =>
+                val mapValue = toCanonicalKeyNameMap(m)
+                // Extract the target parameter from the MapValue representation of the request content body
+                mapValue.get(CName.toCanonicalName(arg.name)) match {
+                  case Some(paramValue) =>
+                    // {"(param name)":(value)}
+                    Some(argCodec.fromMsgPack(paramValue.toMsgpack))
+                  case None =>
+                    // When the target parameter is not found in the MapValue, try mapping the content body as a whole
+                    argCodec.unpackMsgPack(msgpack)
```
takezoe, line 179: Though I know this is an original behavior of the parameter mapping in airframe-http, this behavior is sometimes convenient to take whole request body as a string, however, it might be a little dangerous because this happens unintentionally when we give a wrong parameter name.
```scala
def rpc1(name: String) = {}

{"name": "hello"} => "hello"
{"meme": "hello"} => """{"meme": "hello"}"""
```
If we need to do this intentionally, it would be better to use HttpRequest injection than relying on this fallback behavior.

airframe-http/.jvm/src/test/scala/wvlet/airframe/http/HttpRequestMapperTest.scala
```diff
@@ -0,0 +1,213 @@
...
+  *
+ */
+object HttpRequestMapperTest extends AirSpec {
+
+  case class NestedRequest(name: String, msg: String)
+  case class NestedRequest2(id: Int)
+
+  @RPC
+  trait MyApi {
+    def rpc1(p1: String): Unit = {}
```
takezoe, line 33: By the way, how does airframe-http work if we define the same name methods like:
```scala
def rpc1(p1: String): Unit = {}
def rpc1(p1: String, p1: Int): Unit = {}
```

takezoe:
I saw an unintuitive case in Array (and Map) conversion. 
```scala
def test(p1: Int) = {}
{"p1": [1]} => Seq(null)

def test(p1: Option[Int]) = {}
{"p1": [1]} => Seq(Some(null))
```
I think it's better to throw an exception in these cases, or use the default value (0 in the first case and None in the second case) at least.

takezoe:
In the mapping from the query string, Seq doesn't work for unary parameter while it works for a nested object.
```scala
// Query string: ?name=aaaa&name=bbbb

// This doesn't work
def test(name: Seq[String]) = {}

// This works
case class NestedObject(name: Seq[String])
def test(p1: NestedObject) = {}
```
Even in the second case, if the query string contains a single parameter as `?name=aaaa`, it doesn't work. I wish that Seq("aaaa") is generated in this case.

airframe-http/.jvm/src/test/scala/wvlet/airframe/http/HttpRequestMapperTest.scala
xerial: It will throw a runtime exception when building a Router because these methods will be detected as non-unique mappings. 

I think we can generate different HTTP path mappings for the same name methods. I'll file a ticket.

xerial:
@takezoe `Query string: ?name=aaaa&name=bbbb` should be mapped to Seq[String]. I'll add test cases

airframe-http/.jvm/src/main/scala/wvlet/airframe/http/router/HttpRequestMapper.scala
xerial: Yes. That's an issue and I'm worrying the compatibility of existing applications using airframe-http. For applications using sbt-airframe (e.g., Conductor), we can fix both of the client and server code at the same time, so we don't need to worry about the compatibility.

An idea is totally deprecating the whole content body mapping as you suggested, and ask users to use Request object.

airframe-http/.jvm/src/test/scala/wvlet/airframe/http/HttpRequestMapperTest.scala
xerial: Filed https://github.com/wvlet/airframe/issues/1128

airframe-http/.jvm/src/main/scala/wvlet/airframe/http/router/HttpRequestMapper.scala
xerial: Filed this as https://github.com/wvlet/airframe/issues/1129. At least for RPC, which uses generated clients, we don't need to support the whole body mapping

xerial:
@takezoe Fixed the corner cases mentioned in the above and filed #1128 and #1129 as future work

takezoe:
Thanks for fixing! Please allow me to ask about another corner case.

How should it work `Seq` mapping from the query string if the query string has no corresponding parameter?
```scala
def endpoint3(p1: Seq[String]): Unit = {} // => IllegalArgumentException
def endpoint3(p1: Option[Seq[String]]): Unit = {} // => IllegalArgumentException
```
It would be nice if the first case will be `Nil` and the second case will be `None` as same as the mapping from JSON.

xerial:
@takezoe Right It should bind Seq.empty and None respectively. Added test cases and found how to properly handle these cases (primitive, Seq[primitive], Option[primitive], Option[Seq[primitive]]). Thanks for the comment.

xerial:
Awesome. Your review was really helpful. I appreciate it.

Issue #1131, opened 2020-06-13
airframe-http: Add client interface for using HttpMessage.Request/Response


xerial:
For supporting backend independent http request/response handling. 

```scala
val client = Http.client.withBackend(FinagleClientBackend).newClient("localhost:8080")
client.send(Http.GET("/v1/info")) // => Scala.Future[Response] 
```
If we implement a backend using HttpURLConneciton like sttp https://github.com/softwaremill/sttp/blob/master/core/src/main/scalajvm/sttp/client/HttpURLConnectionBackend.scala, airframe-http can be used with minimum dependencies.

xerial:
Http.SyncClient, Http.AsyncClient is added in #1134

PR #1134, opened 2020-06-14
airframe-http: #1131 Add a sync client using URLConnection 

xerial:
This PR is for adding a handy http client implementation based on java.net.URLConnection, which doesn't require extra dependencies.

- [x] Add Http client implementation using URLConnection
- [x] Add URLConnection specific exception handling
- [x] Add test cases
- [x] Add factory methods
- [x] Add http client documentations

If we can use jdk11, https://openjdk.java.net/groups/net/httpclient/intro.html seems better.

### DIFF
--- a/airframe-control/src/main/scala/wvlet/airframe/control/IO.scala
+++ b/airframe-control/src/main/scala/wvlet/airframe/control/IO.scala
@@ -1,0 +1,60 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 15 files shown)


xerial:
@takezoe PTAL.  Although URLConnection is handy to use, a catch was it doesn't support PATCH (which is used in some of our Operate API method). 

If supporting only JDK11 is acceptable, we should support JDK11's HTTP client backend (HttpClientBackend is an abstraction for adding such implementation later)

takezoe:
Spark related things will need Java8 support for a while. However, using JDK11's HTTP client makes sense. We can build full-featured Airframe HTTP client implementation on it.

airframe-control/src/main/scala/wvlet/airframe/control/IO.scala
```diff
@@ -0,0 +1,60 @@
...
+
+  def readAsString(url: URL): String = {
+    withResource(url.openStream()) { in => readAsString(in) }
+  }
+
+  def readAsString(in: InputStream): String = {
+    readFully(in) { data => new String(data, StandardCharsets.UTF_8) }
+  }
+
+  def readFully[U](in: InputStream)(f: Array[Byte] => U): U = {
```
takezoe, line 38: Does this method need to take a function? How about simply returning `Array[Byte]`?

xerial: That's a good idea. I'll fix it.

PR #1141, opened 2020-06-17
Support Metals + Bloop build

xerial:
- Use `provided` scope for referencing source-level dependencies as bloop doesn't support `compile-internal` scope. compile-internal scope is a special scope of sbt, which will be used for compile time, but not included in pom.xml
- As a workaround, excluded `provided` dependencies from pom.xml using pomPostProcess task, because we will not publish airframe-di-macros (airframe's macros), airspec-deps,core,logs (source code level dependency to airspec) to Maven central, so having provided scope dependency will cause download errors


xerial:
This is FYI. With this change, airframe can be developed with vscode + Metals @takezoe @shimamoto @Pritesh-Patel

PR #1142, opened 2020-06-17
airframe-http: Various client fixes

xerial:
- [x] Use the default design builder in Finagle examples
- [x] Resolve ReflectObjectFactory.isStatic early because sbt may unload classloader early and causes NPE for Thread.getContextClassLoader. This often happened when testing examples project.
- [x] Fix HttpResposneCodec to properly read raw MsgPack 
- [x] Use custom codec properly in FinagleClient
- [x] Throw HttpClientException with 200 code when the request succeeded but converting the response body into the target object failed

### DIFF
--- a/airframe-http/.jvm/src/main/scala/wvlet/airframe/http/router/HttpResponseCodec.scala
+++ b/airframe-http/.jvm/src/main/scala/wvlet/airframe/http/router/HttpResponseCodec.scala
@@ -23,8 +23,8 @@ class HttpResponseCodec[Resp: HttpResponseAdapter] extends MessageCodec[HttpResp
   override def pack(p: Packer, v: HttpResponse[_]): Unit = {
     v.contentType match {
       case Some("application/x-msgpack") =>
+        // Raw msgpack response
         val b = v.contentBytes
-        p.packArrayHeader(b.length)
         p.writePayload(b)
       case Some(x) if x.startsWith("application/json") =>
         // JSON -> MsgPack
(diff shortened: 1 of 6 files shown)


airframe-http/.jvm/src/main/scala/wvlet/airframe/http/router/HttpResponseCodec.scala
```diff
@@ -23,8 +23,8 @@ class HttpResponseCodec[Resp: HttpResponseAdapter] extends MessageCodec[HttpResp
   override def pack(p: Packer, v: HttpResponse[_]): Unit = {
     v.contentType match {
       case Some("application/x-msgpack") =>
+        // Raw msgpack response
```
xerial, line 26: This code path has never been used in our application unless the API returns raw HTTP Response class with msgpack body content.

airframe-http/.jvm/src/main/scala/wvlet/airframe/http/router/HttpResponseCodec.scala
```diff
@@ -23,8 +23,8 @@ class HttpResponseCodec[Resp: HttpResponseAdapter] extends MessageCodec[HttpResp
...
         val b = v.contentBytes
-        p.packArrayHeader(b.length)
```
shimamoto, line 27: No need to call `packBinaryHeader`?

xerial: In this case, the server generates a raw message pack value (It can be MapValue, StringValue, etc.), so returning the byte array (= msgpack) as is.

PR #1143, opened 2020-06-17
airframe-http: #1129 Deprecate no-key unary mapping for RPC

### LINKED ISSUE #1129: airframe-http: Deprecate whole content-body mapping for RPC
The current behavior:
```scala
def rpc1(name: String) = {}

case 1: 
content {"name": "hello"} => name: "hello"

case 2:
content {"meme": "hello"} => name: """{"meme": "hello"}"""
```

The expected behavior:
```scala
case 1: 
content {"name": "hello"} => name: "hello" // OK

case 2:
content {"meme": "hello"} => throw 400 (Bad Request)
```

Related to #766

xerial:
This PR will force using parameter names as keys of RPC call request body. 

For example, if the rpc method is `def f(p1:Int, p2:ComplexObj)`, the request body always needs to have param name -> param value structure:
```
{"p1":1, "p2":{..}}
```

- This will fix unexpected RPC mapping reported in #1129 and will make the request body generation protocol more consistent. 
- sbt-airframe automatically generates such request body, so no client code change will be required. Just need to regenerate RPC ServiceClient with `airframeHttpReload` command. 
- This PR indicates that if you modify RPC method argument names, it will change the RPC request body structure too. So when designing public APIs with airframe-rpc, you need to be careful about the parameter names and their changes. This will be a problem if a client generates an HTTP client based on an older version of the RPC interface (some parameter names will not match between client and RPC server).

This will resolve #1129 and #1100

### DIFF
--- a/airframe-http/.jvm/src/main/scala/wvlet/airframe/http/Router.scala
+++ b/airframe-http/.jvm/src/main/scala/wvlet/airframe/http/Router.scala
@@ -146,7 +146,8 @@ case class Router(
                   controllerSurface,
                   endPoint.method(),
                   prefixPath + endPoint.path(),
-                  m
+                  m,
+                  isRPC = false
                 )
             }
         case (None, Some(rpc)) =>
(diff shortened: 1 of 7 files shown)


xerial:
@takezoe @shimamoto This change limits the impact only for RPC server/client. Let me merge this.

PR #1153, opened 2020-06-22
airframe-di: Add .afterStart lifecycle hook

### LINKED ISSUE #1145: airframe-di: Add more lifecycle hook phases beforeStart, afterStart
The current behavior of onStart, onShutdown is overwriting the existing lifecycle hooks. We occasionally need to add additional lifecycle steps based on the previous design. For example, airframe-http-recorder servers usually have complex initialization steps for adding test records. If we want to add more records and share that initialization steps using Design, we need to append one more onStart lifecycle. 

```scala
val mockServerDesign = newDesign
  .bind[HttpRecorder].toInstance(HttpRecorder.createProgrammableServer(...))
  .onStart(x => x.record(r1); x.record(r2)) 

// Application 
mockServerDesign
  .bind[HttpRecorder].onStart(x => x.record(r3))   // r1, r2 will be lost
```

Two approaches:
1) Instead of overwriting lifecycle hooks, onXXX will append a new lifecycle hook (This will be a bad idea because of the behavior change)
2) Or adding an additional lifecycle step like .afterStart { ... } to distinguish the usage with .onStart {...}. onStart is mainly used for a common design, and afterStart is for customization. For example, introducing two steps around start lifecycle:
```
.beforeStart { ... } // new
.onStart { ... } 
.afterStart { ... } // new
```
Similarly,
```
.beforeShutdown(...) // already exists
.onShutdown(...)  // already exists
.afterShutdown(...) // new
```

The behavior of 1) will be difficult to understand and managing Design can be complicated, so 2) looks better (adding before/after-Start/Shutdown hooks)

```Scala
// Application 
mockServerDesign
  .bind[HttpRecorder].afterStart(x => x.record(r3))   // r1, r2, and r3 can be used
```

@takezoe @shimamoto What do you think?

xerial:
Closes #1145

### DIFF
--- a/airframe/src/main/scala/wvlet/airframe/lifecycle/LifeCycleManager.scala
+++ b/airframe/src/main/scala/wvlet/airframe/lifecycle/LifeCycleManager.scala
@@ -185,7 +185,7 @@ class LifeCycleManager(
     }
   }
 
-  def addAfterStartHook(h: LifeCycleHook): Unit = {
+  private def addAfterStartHook(h: LifeCycleHook): Unit = {
     findLifeCycleManagerFor(h.surface) { l =>
       l.synchronized {
         if (l.afterStartHookHolder.registerOnlyOnce(h)) {
--- a/airframe/src/main/scala/wvlet/airframe/lifecycle/LifeCycleManager.scala
+++ b/airframe/src/main/scala/wvlet/airframe/lifecycle/LifeCycleManager.scala
@@ -34,6 +34,7 @@ sealed trait LifeCycleHookType
 case object ON_INIT         extends LifeCycleHookType
 case object ON_INJECT       extends LifeCycleHookType
 case object ON_START        extends LifeCycleHookType
+case object AFTER_START     extends LifeCycleHookType
 case object BEFORE_SHUTDOWN extends LifeCycleHookType
 case object ON_SHUTDOWN     extends LifeCycleHookType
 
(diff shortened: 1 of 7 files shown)


xerial:
@takezoe Let me add this afterStart hook support so that we can add more lifecycle hook to the design which already has onStart hook.

airframe/src/main/scala/wvlet/airframe/lifecycle/LifeCycleManager.scala
```diff
@@ -310,6 +332,13 @@ object FILOLifeCycleHookExecutor extends LifeCycleEventHandler with LogSupport {
     }
   }
 
+  override def afterStart(lifeCycleManager: LifeCycleManager): Unit = {
```
takezoe, line 335: I guess `afterStart` comes from after `onStart`. However, it's hard to choose an appropriate method by considering which methods are already used (e.g. I have to use `afterStart` because `onStart` is already used). Rather, it might be better to give a name which expresses the purpose of the method. In this case, a name which expresses that this method should be used for customizing a component. I thought `onInit` is better for customization, but it already exists... 😓 

Or, just explaining the purpose of this life cycle method on Scaladoc or README might be enough.

xerial: It's a valid concern, so I'd like to describe the mode details about the appropriate use case of  `afterStart` (and `beforeShutdown`).

The lifecycle phases are:
onInit (instance initialization) -> onInject (when DI is injecting instance, can be called multiple times) -> onStart (when session starts) -> afterStart (after onStart, use this only for customizing test design) -> beforeShutdown -> onShutdown 

Currently, `.onInit` is used for _instance_ initialization _before_ the injection regardless of the session state (not started or started).

PR #1170, opened 2020-06-30
Add SocketTimeoutException as a retryable failure

### LINKED ISSUE #1168: airframe-http-okhttp: Handle IOException
cc: @shimamoto

shimamoto:
Closes #1168 

Add the following exceptions as retryable failures.

- java.net.SocketTimeoutException
- java.net.SocketException and the message is "Socket closed"

### DIFF
--- a/airframe-http-okhttp/src/test/scala/wvlet/airframe/http/okhttp/OkHttpClientTest.scala
+++ b/airframe-http-okhttp/src/test/scala/wvlet/airframe/http/okhttp/OkHttpClientTest.scala
@@ -1,5 +1,7 @@
 package wvlet.airframe.http.okhttp
 
+import java.time.Duration
+
 import wvlet.airframe.control.Control.withResource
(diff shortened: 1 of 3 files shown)


airframe-http-okhttp/src/test/scala/wvlet/airframe/http/okhttp/OkHttpClientTest.scala
```diff
@@ -77,6 +79,12 @@ trait FinagleClientTestApi extends LogSupport {
     Response(Status.Forbidden)
   }
 
+  @Endpoint(method = HttpMethod.GET, path = "/readtimeout")
+  def readTimeout: String = {
+    Thread.sleep(3000)
```
xerial, line 84: ok. So it doesn't mean we always wait 3 sec x 2 (with retry) because timeout is configured to 10ms. Good

PR #1184, opened 2020-07-16
Support a Date instance as an argument for withDate method

yuokada:


airframe-http/src/main/scala/wvlet/airframe/http/HttpMessage.scala
```diff
@@ -114,11 +116,12 @@ trait HttpMessage[Raw] {
   def withContentTypeMsgPack: Raw                       = withContentType(HttpHeader.MediaType.ApplicationMsgPack)
   def withContentLength(length: Long): Raw              = withHeader(HttpHeader.ContentLength, length.toString)
   def withDate(date: String): Raw                       = withHeader(HttpHeader.Date, date)
+  def withDate(date: Date): Raw                         = withHeader(HttpHeader.Date, FinagleMessage.httpDateFormat(date))
```
yuokada, line 119: The implementation for `Message. httpDateFormat` is so small that we should implement this feature without `finagle-base-http`.

https://github.com/twitter/finagle/blob/34167989a6dc964e24e20212516fba89fa58bb23/finagle-base-http/src/main/scala/com/twitter/finagle/http/Message.scala#L642-L658

xerial: ZonedDateTime is not supported in Scala.js, so we need to use Instant, which is implemented in https://github.com/scala-js/scala-js-java-time
. 

I'm not sure about java.util.Date. I guess this is not supported too. 

It's better to make it an extension only for JVM project, using some implicit class

xerial: For example, Date time is not supported in Scala.js, so airframe-codec supports Date and ZonedDateTime only for JVM projects https://github.com/wvlet/airframe/blob/master/airframe-codec/.jvm/src/main/scala/wvlet/airframe/codec/JavaTimeCodec.scala

xerial: Two options:
- Use withDate(instant) 
- Add an implicit class RichMessage(val message:HttpMessage) extends AnyVal { def withDate(...) } only for airframe-http (JVM)

I think the first one is easier

build.sbt
```diff
@@ -564,7 +564,10 @@ lazy val http =
     .settings(buildSettings)
     .settings(
       name := "airframe-http",
-      description := "REST API Framework"
+      description := "REST API Framework",
+      libraryDependencies ++= Seq(
+        "com.twitter" %% "finagle-base-http"   % FINAGLE_VERSION,
```
xerial, line 569: airframe-http should not depend on finagle to minimize the dependencies. That's why we are isolating airframe-http-finagle as another project.

airframe-http/src/main/scala/wvlet/airframe/http/HttpMessage.scala
xerial: The third option: implement Scala JVM/Scala.js specific httpDateFormatter using Compat.scala files

xerial:
@yuokada Any update? Let me know if you need a help.

xerial:
Okay. It seems compiling this code in Scala.js has no problem. Let me merge this. Thanks!

PR #1192, opened 2020-07-26
airframe-grpc: Support gRPC backend

xerial:
Added an experimental implementation of gRPC backend:
- [x] Use airframe-codec for gRPC request/response serialization
- [x] Build gRPC server based on Airframe RPC router
- [x] Add gRPC server factory

TODO in another PR:
- [ ] Generate gRPC client stub wiht sbt-airframe plugin.
- [ ] Support client, server-side, and bidirectional streaming
- [ ] Add a gRPC server proxy with airframe-http-finagle for supporting HTTP1


docs/airframe-rpc.md
```diff
@@ -374,6 +374,46 @@ trait MyAPI {
...
+## Airframe gRPC
+
+_(This is an experimental feature available since Airframe 20.8.0)_
+
+Airframe gRPC is a gRPC and HTTP2-based implementation of Airframe RPC, which can make thousands of RPC calls per second. With Airframe gRPC:   
+
+- No Protobuf definition is required. You can use plain Scala and case classes to define gRPC service.
+- Roadmap
+  - [x] Create a gRPC server from Airframe RPC router
+  - [ ] Generate gRPC client stub wiht sbt-airframe plugin.
```
xuwei-k, line 387:
```suggestion
  - [ ] Generate gRPC client stub with sbt-airframe plugin.
```

xerial: thx!

xerial:
Good point!

The initial motivation is leveraging HTTP2 and server/client-side streaming of gRPC so that we can build fast and interactive RPC services. 

Generating a Protobuf IDL from an Airframe RPC interface looks promising and should be easy. What we need to do will be adding a filter for converting Protobuf requests into MessagePack requests to call RPC methods. Then we can leverage existing gRPC client generators for various programming languages. I think most of the people who are already familiar with gRPC will be interested in this direction (Airframe RPC -> Protobuf IDL -> Protobuf gRPC client). 

For Scala/Java users, we need to provide sbt-airframe Airframe gRPC client generator, which should be the most efficient client implementation.

PR #1194, opened 2020-07-28
airframe-grpc: Genearte gRPC clients

xerial:
- Add gRPC client code generator
  - [x] blocking client (grpc-sync)
  - non-blocking client (grpc-async)  (todo in another PR)
- [x] Add sbt-airframe test for gRPC
- [x] Update airframe-rpc.md about the client generation

### DIFF
--- a/docs/airframe-rpc.md
+++ b/docs/airframe-rpc.md
@@ -398,12 +398,12 @@ __build.sbt__
 
 ```scala
 import wvlet.airframe.http.Router
-import wvlet.airframe.http.grpc.Grpc
+import wvlet.airframe.http.grpc.gRPC
 
 // Create a Router definition in the same manner with Airframe RPC
 val router = Router.add[MyApiImpl]
 
-val grpcServerDesign = Grpc.server
+val grpcServerDesign = gRPC.server
   .withRouter(router)
   .withPort(8080)
   .design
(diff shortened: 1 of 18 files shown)


sbt-airframe/src/sbt-test/sbt-airframe/generate-client/server/src/main/scala/myapp/server/MyServer.scala
```diff
@@ -31,36 +31,36 @@ object MyServer extends LogSupport {
 
     d.build[FinagleClient] { finagleClient =>
       val client = new myapp.spi.ServiceClient(finagleClient)
-      val future = client.myService.hello(100).map { v =>
+      val future = client.MyService.hello(100).map { v =>
```
xerial, line 34: This will be a breaking change for existing http clients generated by sbt-airframe (cc: @takezoe @shimamoto @Pritesh-Patel )

docs/airframe-rpc.md
```diff
@@ -210,6 +210,7 @@ Supported client types are:
 - __sync__: Create a sync HTTP client (ServiceSyncClient) for Scala (JVM)
 - __async__: Create an async HTTP client (ServiceClient) for Scala (JVM) using Future abstraction (`F`). The `F` can be `scala.concurrent.Future` or twitter-util's Future. 
 - __scalajs__:  Create an RPC client (ServiceClientJS)
+- __grpc-sync__: Create a gRPC blocking client (ServiceGrpcSyncClient)
```
xerial, line 213: This option is subject to change because generating sync, async, and Future-based clients inside the same class would be possible.

sbt-airframe/src/sbt-test/sbt-airframe/generate-client/server/src/main/scala/myapp/server/MyServer.scala
xerial: Yes. For us, conductor's Scala.js code will need to be fixed after upgrading to airframe 20.8.0.

xerial:
@shimamoto For retry, I think we can use grpc-java's default retry, which can be set at the ManagedChannel, or we will need to add an extension point for adding custom grpc intercetor. That's a good point. Let me try that after merging this PR.

PR #1196, opened 2020-07-29
Shutdown OkHttpClient explicitly

takezoe:
According to OkHttp document, we don't need to shutdown OkHttp explicitly.
https://square.github.io/okhttp/4.x/okhttp/okhttp3/-ok-http-client/#shutdown-isnt-necessary

However, particularly, in CLI applications, because OkHttp threads remain (during max-idle period?) the process doesn't exit even after execution without explicit `System.exit(0)` call. It would better to shutdown explicitly in lifecycle hook.

### DIFF
--- a/airframe-http-okhttp/src/main/scala/wvlet/airframe/http/okhttp/OkHttpClient.scala
+++ b/airframe-http-okhttp/src/main/scala/wvlet/airframe/http/okhttp/OkHttpClient.scala
@@ -242,7 +242,10 @@ class OkHttpClient(address: ServerAddress, config: OkHttpClientConfig)
     convert[OperationResponse](send(r, requestFilter))
   }
 
-  def close(): Unit = {}
+  def close(): Unit = {
+    client.dispatcher().executorService().shutdown()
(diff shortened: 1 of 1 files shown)


airframe-http-okhttp/src/main/scala/wvlet/airframe/http/okhttp/OkHttpClient.scala
```diff
@@ -242,7 +242,10 @@ class OkHttpClient(address: ServerAddress, config: OkHttpClientConfig)
     convert[OperationResponse](send(r, requestFilter))
   }
 
-  def close(): Unit = {}
+  def close(): Unit = {
```
xerial, line 245: ok. Then, let's implement AutoCloseable and override close() method

takezoe: `HttpSyncClient` already has `AutoCloseable`, but we hasn't gave `override` modifier to `close()` method!

xerial:
Thanks. Merged

PR #1205, opened 2020-08-06
Fix typo

xirc:


xirc:
I fixed some typos.

By the way, I found two README files for _airframe-di_.
Which is source of truth? both?

xerial:
Thanks! docs/airframe.md will be the source of truth eventually. The airframe/README.md should have a pointer to the docs/airframe.md

PR #1214, opened 2020-08-08
airframe-rx: Add RxOption[A] 

### LINKED ISSUE #1200: NPE at RxVar.update  in Scala.js
```scala
    def update(updater: A => A, force: Boolean = false): Unit = {
      val newValue = updater(currentValue)
      if (force || currentValue != newValue) {
        currentValue = newValue
        subscribers.map { s => s(newValue) } <-- 
      }
    }
```

xerial:
This is for simplifying Rx[Option[A]] handling. Currently, we need to nest multiple map calls:
```scala
val a = Rx.variable(Some("hello")) 
a.map { opt =>
   opt.map { x => div(s"message: ${x}") }
} // Rx[Option[RxElement]]
```

This code will be simpler like:
```scala
val a: RxOptionVar[A] = Rx.optionVariable(Some("hello"))
a.map { x => div(s" message: ${x}") } // Rx[Option[RxElement]]

a := Some("new value")
```

This PR also supports Rx[A].toOption if A is Option[_] type 

This also closes #1200

### DIFF
--- a/airframe-http-rx/src/main/scala/wvlet/airframe/http/rx/Rx.scala
+++ b/airframe-http-rx/src/main/scala/wvlet/airframe/http/rx/Rx.scala
@@ -51,7 +51,7 @@ object Rx extends LogSupport {
 
   def apply[A](v: A): RxVar[A]                        = variable(v)
   def variable[A](v: A): RxVar[A]                     = new RxVar(v)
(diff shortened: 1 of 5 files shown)


xerial:
@shimamoto Let me merge this PR as it's becoming a blocker for supporting gRPC streaming #1226 #1227

PR #1220, opened 2020-08-09
airfame-codec: Throw an exception for invalid enum string

### LINKED ISSUE #1201: airframe-codec: MessageCodec.fromXXX should throw an exception for invalid values
StringUnapplyCodec.fromString("xxx") will not throw any exception, but returns null for Enum-like classes, which have unapply(str:String): Option[X] method. 

This behavior should be changed so that we can check invalid values.

xerial:
Closes #1201


xerial:
@shimamoto Could you take a look? This fix will be necessary to serialize our internal case classes like Stage, Site, etc.

PR #1225, opened 2020-08-11
Add JSON.format for pretty printing

takezoe:


airframe-json/src/main/scala/wvlet/airframe/json/JSON.scala
```diff
@@ -65,6 +65,48 @@ object JSON extends LogSupport {
     JSONScanner.scanAny(s, b)
   }
 
+  /**
+    * Format JSON value
+    */
+  def format(v: JSONValue): String = {
```
xerial, line 71: To control spacing (e.g., spaceBeforeArray, Object, line width, etc.), we will need to add JSONFormatConfig as an optional second argument.

takezoe: Making formatting configurable sounds nice! Further options can be considerable. For example, `wrapArrayElement` which works as:
```javascript
// wrapArrayElement = false
[1, 2, 3]

// wrapArrayElement = true (current behavior)
[
  1,
  2,
  3,
]
```

Issue #1227, opened 2020-08-11
airframe-rx: RxStream[X]


xerial:
Adding a streaming interface for supporting gRPC streaming #1226 

```scala
@RPC
trait MyApi {
  def unaryCall(args...): ReturnType
  def serverStreaming(args...): RxStream[ReturnType]
  def clientStreaming(args..., input:RxStream[RequestType]): ReturnType
  def bidiStreaming(args..., input:RxStream[RequestType]): RxStream[ReturnType]
}
```

With this approach, we don't need to generate ServerImpl stub as we can use RxStream[X] type as is. 

- [x] Migrate the code in airframe-http-rx into wvlet.airframe.http.rx into airframe-http so that we can use it for defining RPC interface
- [x] Add monadic operators to RxStream (e.g., map, filter(...), buffer(...))
- [ ] Wrap grpc-java's StreamObserver with RxStream

xerial:
To implement this, Rx.run needs to handle three states: onNext, onError, and onCompletion (Observable interface). Without managing these states, detecting the end of streams, which is necessary for calling StreamObserver.onError/onCompletion, cannot be handled.

xerial:
Now that airframe-rx natively supports error and completion event propagation, we don't need to add a new interface. We can just use Rx[X] https://wvlet.org/airframe/docs/airframe-rx

PR #1228, opened 2020-08-12
airframe-rx: Move Rx classes to airframe-http project

xerial:
This change is for adding a reactive stream interface to airframe-http, which will be used for gRPC #1227 

DOM related classes (e.g., RxElement, HTML tags, etc.) will remain in airframe-http-rx, so there will be no breaking changes for the library users.


xerial:
cc: @shimamoto

PR #1233, opened 2020-08-13
Replace a body's source with a byte array in memory

### LINKED ISSUE #1212: airframe-okhttp: Need to close ResponseBody
OkHttpResponseAdapter and OkHttpClient implementation need to be redesigned to close ResponseBody, which can be only read once, appropriately. 
```
warn [OkHttpClient] A connection to https://xxxxx was leaked. Did you forget to close a response body? To see where this was allocated, set the OkHttpClient logger level to FINE: Logger.getLogger(OkHttpClient.class.getName()).setLevel(Level.FINE);
```
cc: @shimamoto @takezoe

shimamoto:
okhttp has modeled responses to be one-shot. They expect us to stream the response, decode it from JSON or another format, and use the result as a model object. This PR is to replace the streaming one with the in-memory one so that we can pass the response along, which is one of the options.

How about this solution?

Closes #1212

### DIFF
--- a/airframe-http-okhttp/src/main/scala/wvlet/airframe/http/okhttp/OkHttpRetryInterceptor.scala
+++ b/airframe-http-okhttp/src/main/scala/wvlet/airframe/http/okhttp/OkHttpRetryInterceptor.scala
@@ -1,6 +1,6 @@
 package wvlet.airframe.http.okhttp
 
-import okhttp3.{Interceptor, Response}
+import okhttp3.{Interceptor, Response, ResponseBody}
 import wvlet.airframe.control.ResultClass
 import wvlet.airframe.control.Retry.RetryContext
 import wvlet.airframe.http.HttpClientMaxRetryException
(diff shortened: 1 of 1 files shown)


airframe-http-okhttp/src/main/scala/wvlet/airframe/http/okhttp/OkHttpRetryInterceptor.scala
```diff
@@ -47,6 +47,21 @@ class OkHttpRetryInterceptor(retry: RetryContext) extends Interceptor {
     }
   }
 
+  private def inMemoryResponse(response: Response) = {
+    Option(response.body)
+      .map { body =>
+        // The only stateful object is a body's source, which is a one-shot value and then must be closed.
+        // Replace it with a byte array in memory so that we can pass the response along.
+        val r = response
+          .newBuilder().body(ResponseBody.create(body.contentType, Try(body.bytes).getOrElse(Array.empty))).build()
```
xerial, line 56: Does it copy the response headers the new response instance?

xerial: ok. It looks like so.

xerial:
OK. Let's try this implementation. For the current use cases, we don't expect reading large responses with okhttp.

PR #1235, opened 2020-08-14
airframe-http-rx: Represent Rx values with OnNext, OnError, and OnCompletion

xerial:
This change will be necessary for supporting gRPC streaming #1234 

In this PR, Rx values will be represented with three states: OnNext(v), OnError(Throwable), and OnCompletion. Especially OnCompletion event is important to implement Rx.lastOption, Rx.concat, Rx.sequence. Without OnCompletion event, we don't know the end of stream (lastOption), so we can't concatenate two or more streams. 

With this change, Rx can be used as a reactive programming interface http://reactivex.io/

### DIFF
--- a/airframe-http-rx/.js/src/main/scala/wvlet/airframe/http/rx/html/DOMRenderer.scala
+++ b/airframe-http-rx/.js/src/main/scala/wvlet/airframe/http/rx/html/DOMRenderer.scala
@@ -14,10 +14,11 @@
 package wvlet.airframe.http.rx.html
 
 import org.scalajs.dom
-import wvlet.airframe.http.rx.{Cancelable, Rx}
+import wvlet.airframe.http.rx.{Cancelable, OnCompletion, OnError, OnNext, Rx, RxRunner}
 import wvlet.log.LogSupport
 
 import scala.scalajs.js
+import scala.util.Try
 
 /**
   * Convert HtmlNodes into DOM elements for Scala.js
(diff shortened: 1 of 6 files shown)


xerial:
@shimamoto @takezoe @yuokada 

I think reviewing this PR is not so easy, so let me explain the background of this change.

This Rx interface is similar to scala.concurrent.Future, com.twitter.util.Future, or monix. Generally speaking, it's an implementation of http://reactivex.io/. 

Rx[A] is a chain of operators that will be applied to the stream of objects A and it's used for dynamically updating a part of DOM elements in airframe-http-rx. See DOMRenderer code for more details.

This Rx implementation was originally created by following monadic-html https://github.com/OlivierBlanvillain/monadic-html, but this has no error handling and error propagation to the upstream operators, so reporting OnError(exception) or OnCompletion event was difficult in monadic-html. Knowing the end of streams is important to support gRPC, which needs to terminate RPC connection at some point, and also for combining multiple streams (e.g., concat, join, zip, etc.) 

With this PR, adding more ReactiveX methods will also be possible (e.g., Rx.buffer, Rx.retry, etc.). And Rx operator case classes are independent from RxRunner, so it is also possible to add different RxRunners for supporting concurrent programming if we want. 

For our use cases, we only need reactive DOM rendering (RxElement) and gRPC streaming. After merging this PR, I'd like to extract this code as airframe-rx module under wvlet.airframe.rx package to maintain this code as a thin reactive interface with less dependencies.

PR #1238, opened 2020-08-17
airframe-rx-html: Relocate to wvlet.airframe.rx.html

xerial:
Moving RxElement to wvlet.airframe.rx.html package.

This is because RxElement has no dependency to airframe-http now that airframe-rx has moved as an independent module.


xerial:
FYI @shimamoto @takezoe

PR #1239, opened 2020-08-17
airframe-grpc: Support client, server, and bidirectional streaming

### LINKED ISSUE #1226: airframe-grpc: Streaming interface
Design questions:
- Should we use StreamObserver?  It has no control for the back-pressure. Related discussion can be found in https://github.com/grpc/grpc-java/issues/1549 
- What interface should we use? Stream[X] is already in Scala collection, so using it can be confusing.  => We can use `Rx[X]` 
- grpc-kotlin is using Flow with coroutine 
- Using scala-async can be an option? 

Related:
- https://github.com/salesforce/reactive-grpc
- https://github.com/fiadliel/fs2-grpc

xerial:
This will close #1226 

- [x] Use Rx[X] interface for streaming 
- [x] Generate streaming client code
- [x] Add bridges between Rx[X] and StreamObserver[X] of gRPC


xerial:
@takezoe @shimamoto 
This is the initial streaming support for gRPC. The code still has rough edges and lacks some error handling. Let me merge this PR first. I'll improve these things over time.

PR #1244, opened 2020-08-18
sbt-airframe: Limit the scope of finding @RPC annotations

### LINKED ISSUE #1242: abt-airframe: Multiple routes error in multi-projects if impl in placed under the api definiton package
I'm not sure exact conditions to reproduce. However, the keys are:

- Implementation of API is placed in the same or under the package where api trait is placed.
- Generate client in a different sbt sub-project from the above.

This project can reproduce: https://github.com/takezoe/airframe-http-plugin-repro

Of course, there are some workarounds:

- Place implementation in the different parent package.
- Separate api definition and impl to different sub-projects, and client project refers only api project.

However, it would be nice if we can use airframe-rpc-client without such special care.

xerial:
Fixes #1242


takezoe (approved):
Thanks!

PR #1250, opened 2020-08-20
airframe-rpc: Add an RPC project example

xerial:


xerial:
FYI: @wvlet/airframe-dev

PR #1263, opened 2020-09-02
airframe-rx: Fix the compilation error when binding RxVar

xerial:


xerial:
@shimamoto A new test code reproduces the compilation error and making some constructor parameter private fixed this issue. I'll release a new version with this fix.

airframe-rx/src/main/scala/wvlet/airframe/rx/RxVar.scala
```diff
@@ -19,7 +19,7 @@ import scala.language.higherKinds
 /**
   * A reactive variable supporting update and propagation of the updated value to the chained operators
   */
-class RxVar[A](protected var currentValue: A) extends Rx[A] with RxVarOps[A] {
+class RxVar[A](private var currentValue: A) extends Rx[A] with RxVarOps[A] {
```
xerial, line 22: Not sure the exact cause of the compilation error, but I guess airframe-surface or DI might have tried to create an accessor function to the protected variable, and caused the compilation error.

PR #1264, opened 2020-09-02
internal: Add packaging test

xerial:
Added a new CI task to test publish (including package, doc) works properly before releasing artifacts. Without this check, releasing at CI can fail and fixing the issue can be tricky.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -31,7 +31,7 @@ addCommandAlias(
 )
 addCommandAlias(
   "publishJVMLocal",
-  s"; ++ ${SCALA_2_12}; projectJVM2_13/publishLocal; projectJVM2_12/publishLocal; projectJS/publishLocal; sbtAirframe/publishLocal;"
+  s"; ++ ${SCALA_2_12}; projectJVM2_13/publishLocal; projectJVM2_12/publishLocal; sbtAirframe/publishLocal;"
 )
 
 // A workaround for https://github.com/sbt/sbt/issues/5586
--- a/build.sbt
+++ b/build.sbt
@@ -29,6 +29,10 @@ addCommandAlias(
   "publishSnapshots",
   s"; ++ ${SCALA_2_12}; projectJVM2_13/publish; projectJVM2_12/publish; projectJS/publish; sbtAirframe/publish;"
 )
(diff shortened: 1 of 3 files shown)


xerial:
@shimamoto Could you take a look? 20.9.0 release was failed because the doc task (at airframe-benchmark) had a missing dependency. With this test, we can confirm publishing task will work as expected.

build.sbt
```diff
@@ -725,7 +733,8 @@ lazy val benchmark =
         // For ScalaPB
         // "com.thesamet.scalapb" %% "scalapb-runtime-grpc" % scalapb.compiler.Version.scalapbVersion
         // For grpc-java
-        "io.grpc" % "grpc-protobuf" % GRPC_VERSION
+        "io.grpc"             % "grpc-protobuf" % GRPC_VERSION,
+        "com.google.protobuf" % "protobuf-java" % "3.11.0"
```
xerial, line 737: Because of this missing dependency, 20.9.0 release was failed once so I needed to manually publish 20.9.0 artifacts.

build.sbt
```diff
@@ -29,6 +29,10 @@ addCommandAlias(
   "publishSnapshots",
   s"; ++ ${SCALA_2_12}; projectJVM2_13/publish; projectJVM2_12/publish; projectJS/publish; sbtAirframe/publish;"
 )
+addCommandAlias(
+  "publishJVMLocal",
+  s"; ++ ${SCALA_2_12}; projectJVM2_13/publishLocal; projectJVM2_12/publishLocal; projectJS/publishLocal; sbtAirframe/publishLocal;"
```
shimamoto, line 34: publishJSLocal can cover `projectJS/publishLocal`, so looks like projectJS can be removed from publishJVMLocal.

build.sbt
shimamoto: Oh, that must have been tough. Thank you

shimamoto (approved):
Looks good!

build.sbt
xerial: nice catch. Thanks

xerial:
Changed to run packaging test only at master branch to save CI time.

PR #1268, opened 2020-09-04
airframe-rx: Add RxOption.getOrElse/orElse/transform

xerial:


xerial:
@shimamoto This will improve the usability of RxOption and RxOptionVar.

PR #1269, opened 2020-09-05
airframe-rx: Add Rx.take(n) and Rx.interval

xerial:
For periodically monitoring events:
- Interval http://reactivex.io/documentation/operators/interval.html
- Take http://reactivex.io/documentation/operators/take.html

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,4 +32,5 @@ website/i18n/*
 
 **/log/*json*
 .vscode/launch.json
-project/metals.sbt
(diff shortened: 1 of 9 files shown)


xerial:
@shimamoto Based on this implementation, I'll add throttle (debounce) http://reactivex.io/documentation/operators/debounce.html and sample operators http://reactivex.io/documentation/operators/sample.html, which will be useful for monitoring states.

PR #1275, opened 2020-09-09
airframe-rx: Make RxOption public

xerial:
- Add Rx.join/zip/concat convenient methods


xerial:
This is necessary for airframe 20.9.1 cc: @wvlet/airframe-dev

PR #1292, opened 2020-09-16
airframe-rx: Interface hierarchy change: Rx[A], RxOption[A], and RxStream[A]

xerial:
For properly rendering RxOption[RxElements], we need to report `Option[RxElement]` values to the subscriber. Previously it 
 reported RxElement value only when the RxOption value is Some(x), so even if the input value is updated to None, rerendering didn't happen and it left orphaned DOM elements to UI. 

So I needed to change the base type of `RxOption[A]` as `Rx[Option[A]]`. However, `RxOption.map[B](...) : RxOption[B]` and `Rx.map[B](...): Rx[B]` have different return types, so `RxOption[A]` cannot be a subclass of `Rx[Option[A]]`. As a workaround, I separated `Rx[A]` into `RxOption[A]` and `RxStream[A]`. 

With this change, we can properly clear `RxOption[RxElement]` elements (see RxRenderingTest) even if the value is None.

Other changes:
- Rx.fromFuture[A](f:Future[A]) returns RxOption[A] instead of Rx[Option[A]]
- Added Rx.zip4
- Deprecated Rx.apply(...) in favor of Rx.variable(...)
- Added RxOption.transformRx 
- Use RxStream for grpc streaming
- Make Rx.toSeq blocking and changed to support only Scala JVM.


airframe-rx/src/main/scala/wvlet/airframe/rx/Rx.scala
```diff
@@ -155,38 +72,115 @@ trait Rx[+A] extends LogSupport {
...
-  def join[A, B, C, D](a: Rx[A], b: Rx[B], c: Rx[C], d: Rx[D]): Rx[(A, B, C, D)] = Join4Op(a, b, c, d)
+  def join[A, B](a: Rx[A], b: Rx[B]): RxStream[(A, B)]                                 = JoinOp(a, b)
+  def join[A, B, C](a: Rx[A], b: Rx[B], c: Rx[C]): RxStream[(A, B, C)]                 = Join3Op(a, b, c)
+  def join[A, B, C, D](a: Rx[A], b: Rx[B], c: Rx[C], d: Rx[D]): RxStream[(A, B, C, D)] = Join4Op(a, b, c, d)
 
-  def zip[A, B](a: Rx[A], b: Rx[B]): Rx[(A, B)]                 = ZipOp(a, b)
-  def zip[A, B, C](a: Rx[A], b: Rx[B], c: Rx[C]): Rx[(A, B, C)] = Zip3Op(a, b, c)
+  def zip[A, B](a: Rx[A], b: Rx[B]): RxStream[(A, B)]                                 = ZipOp(a, b)
+  def zip[A, B, C](a: Rx[A], b: Rx[B], c: Rx[C]): RxStream[(A, B, C)]                 = Zip3Op(a, b, c)
+  def zip[A, B, C, D](a: Rx[A], b: Rx[B], c: Rx[C], d: Rx[D]): RxStream[(A, B, C, D)] = Zip4Op(a, b, c, d)
```
xerial, line 181: Added zip4 for consistency with join4

xerial:
@shimamoto This will fix a rendering issue when `RxOption.map { ... => div(...) }` is used. div(...) element generated when the value is Some(x) is remaining in the UI after calling cancel() method. With this PR, this issue can be fixed.

I've tested this change in https://github.com/treasure-data/presto-conductor/pull/333.

airframe-rx-html/.js/src/test/scala/wvlet/airframe/http/rx/html/RxRenderingTest.scala
```diff
@@ -0,0 +1,68 @@
...
+        "menu:",
+        title.map { x =>
+          b(x)
+        }
+      )
+    }
+
+    n.outerHTML shouldBe """<div>menu:<b>home</b></div>"""
+    title := None
+    n.outerHTML shouldBe """<div>menu:</div>"""
```
xerial, line 65: @shimamoto FYI. Previously this returned `<div>menu:<b>home</b></div>` and setting None was ignored.

PR #1294, opened 2020-09-16
airframe-rx: Add RxResult for propagating downstream operator states

xerial:
This is only for internal improvements.


xerial:
@shimamoto This is just for internal code readability and preparation when we need to implement backpressure support.

PR #1306, opened 2020-10-05
internal: Upgrade to sbt 1.4.0

xerial:
- Upgrade to sbt 1.4.0
- Enable the build-in dependency tree plugin (We can use dependencyTree command)
- Tested build pipelining https://github.com/sbt/sbt/pull/5703 
, but observed many compilation failures especially around macro / JVM projects. Disabled this feature for now (this behavior is the default anyway)


xerial:
PTAL @shimamoto @takezoe

PR #1310, opened 2020-10-07
internal: Avoid running duplicated GitHub Actions for PR and push

xerial:


xerial:
No. This syntax doesn't work

Issue #1319, opened 2020-10-17
examples/rpc-examples/hello-rpc/ causes a compilation error


NomadBlacky:
I tried the example project of [examples/rpc-examples/hello-rpc/](https://github.com/wvlet/airframe/tree/master/examples/rpc-examples/hello-rpc), but I met a compilation error.

[source code](https://github.com/NomadBlacky/airframe-hello-rpc)

## logs

```
$ sbt
[info] welcome to sbt 1.4.0 (AdoptOpenJDK Java 1.8.0_252)
[info] loading settings for project global-plugins from sbt-updates.sbt ...
[info] loading global plugins from /home/nomadblacky/.sbt/1.0/plugins
[info] loading settings for project airframe-hello-rpc-build from plugins.sbt ...
[info] loading project definition from /home/nomadblacky/ghq/github.com/NomadBlacky/airframe-hello-rpc/project
[warn] There may be incompatibilities among your library dependencies; run 'evicted' to see detailed eviction warnings.
[info] loading settings for project greeter from build.sbt ...
[info] set current project to greeter (in build file:/home/nomadblacky/ghq/github.com/NomadBlacky/airframe-hello-rpc/)
[info]
[info] Here are some highlights of this release:
[info]   - Build server protocol (BSP) support
[info]   - sbtn: a native thin client for sbt
[info]   - VirtualFile + RemoteCache: caches build artifacts across different machines
[info]   - ThisBuild / versionScheme to take the guessing out of eviction warning
[info] See http://eed3si9n.com/sbt-1.4.0 for full release notes.
[info] Hide the banner for this release by running `skipBanner`.
[info] sbt server started at local:///home/nomadblacky/.sbt/1.0/server/5bd55ad33ab16ecbace2/sock
sbt:greeter> compile
2020-10-17 23:21:31.769+0900  info [AirframeHttpPlugin] Extracting airframe-http 20.8.0 package to /home/nomadblacky/ghq/github.com/NomadBlacky/airframe-hello-rpc/target/scala-2.12/airframe/20.8.0/local  - (AirframeHttpPlugin.scala:141)
[warn] There may be incompatibilities among your library dependencies; run 'evicted' to see detailed eviction warnings.
2020-10-17 23:21:33.330+0900 debug [HttpCodeGenerator] HttpClientGeneratorConfig(greeter.api,wvlet.airframe.http.codegen.client.SyncClient$@e36bb2a,greeter.api)  - (HttpCodeGenerator.scala:149)
[...]

PR #1322, opened 2020-10-18
airframe-rpc: Add an RPC demo code using Scala + Scala.js

xerial:


xerial:
Created a new RPC example projects. This also fixes #1319

PR #1347, opened 2020-11-09
Scala 2.11 deprecation

xerial:
- We no longer need to maintain Spark 2.4.x with Scala 2.11 as Spark 3.0.x is stable to use
- This PR drops Scala 2.11 configurations and merges project settings for Scala 2.12 and Scala 2.13 for simplicity
  - Added minor fixes for passing compilation with Scala 2.13

### DIFF
--- a/.github/workflows/release-js.yml
+++ b/.github/workflows/release-js.yml
@@ -4,6 +4,7 @@ on:
   push:
     tags:
       - v*
+  workflow_dispatch:
 
 jobs:
   publish_js:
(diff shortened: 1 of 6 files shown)


.github/workflows/release-js.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     tags:
       - v*
+  workflow_dispatch:
```
xerial, line 7: This option will add a support for manually rerunning workflows. Useful when Sonatype has an issue

build.sbt
```diff
@@ -184,20 +156,15 @@ lazy val communityBuildProjects: Seq[ProjectReference] = Seq(
...
   finagle,
   okhttp,
-  httpRecorder
-)
-
-// JVM projects excluded from Scala 2.13 build
-lazy val jvmProjects2_12: Seq[ProjectReference] = Seq(
+  httpRecorder,
   sql,
-  benchmark,
```
shimamoto, line 200: Is it okay to remove benchmark project from jvmProjects?

build.sbt
```diff
@@ -1,13 +1,11 @@
...
 import xerial.sbt.pack.PackPlugin.publishPackArchiveTgz
 
-val SCALA_2_11 = "2.11.12"
 val SCALA_2_12 = "2.12.12"
 val SCALA_2_13 = "2.13.3"
 
-val untilScala2_12      = SCALA_2_12 :: SCALA_2_11 :: Nil
-val targetScalaVersions = SCALA_2_13 :: untilScala2_12
-val exceptScala2_11     = SCALA_2_13 :: SCALA_2_12 :: Nil
+val untilScala2_12      = SCALA_2_12 :: Nil
```
shimamoto, line 7: Looks like we can remove untilScala2_12.

shimamoto (approved):
Added minor comments, but looks good!

build.sbt
xerial: Good catch. I forgot about adding it. Thanks

PR #1351, opened 2020-11-12
Add setThreadLocal method for the pre-defined key TLS_KEY_SERVER_EXCEPTION

shimamoto:
Like FinagleServer.defaultErrorFilter, I sometimes want to rescue specific exceptions in an own Filter but to record them in an access log. The exceptions rescued convert into a new Future.value, meaning no exceptions are recorded by HttpAccessLogFilter. To record them, this PR is to add a method to set an exception.


xerial (approved):

xerial:
ok. Just renamed the method name.

PR #1380, opened 2020-12-05
Separate AirSpec build 

xerial:
This is a first step to proceed Dotty support. 

With this change, source-level dependencies between Airframe and AirSpec will be removed, which means for building Airframe, we will use a pre-built version of AirSpec (e.g., a previous version).


build.sbt
```diff
@@ -7,7 +7,7 @@ val SCALA_3_0           = "3.0.0-M2"
 val targetScalaVersions = SCALA_2_13 :: SCALA_2_12 :: Nil
 val withDotty           = SCALA_3_0 :: targetScalaVersions
 
-val SCALATEST_VERSION               = "3.0.8"
```
xerial, line 10: We no longer support ScalaTest

xerial:
cc: @wvlet/airframe-dev

PR #1381, opened 2020-12-05
airframe-log: Dotty support

xerial:
- [x] Switch sources between Scala 2 and 3
- [x] Use new Macros
  - Similar example https://github.com/Log4s/log4s/pull/57/files
- [x] Added simple integration tests. To run AirSpec tests with Dotty, we also need to migrate surface, di, metrics. 
- [x] Added a release configuration


xerial:
@takezoe @shimamoto It's the first step for supporting Dotty (Scala 3). Because Scala 3 introduces new macros https://scalacenter.github.io/scala-3-migration-guide/docs/macros/migration-tutorial.html, we need to split code for Scala 2 and 3. For now, it works as expected.

After merging this PR, I'll release airframe-log for Scala 3.0.0-M2.

PR #1405, opened 2020-12-19
airframe-di: Split Scala 2 and Scala 3 source folders

xerial:
This PR is preparation for implementing Airframe DI for Scala 3. All macro codes are moved to XXXImpl trait and moved to scala-2 source folder.


xerial:
I'll start implementing Airframe DI for Scala 3 in another PR cc: @takezoe @shimamoto

PR #1412, opened 2020-12-28
airframe-di: Generating DI code with Scala3 Tree rewriter

xerial:
- Generate `new A with DISupport {} ` for `bind[A]`
- Blocker: https://github.com/lampepfl/dotty/issues/10931


xerial:
This code is no longer necessary as we deprecate in-trait injection from Scala 3 #1649

PR #1448, opened 2021-01-23
airframe-http: Fixes #1446. Resolve MIME type of file names with multiple dots

### LINKED ISSUE #1446: airframe-http: StaticContent failed to set content type
Example: boostrap.min.css is not handled as text/css

xerial:


xerial:
cc: @shimamoto

PR #1450, opened 2021-01-24
airframe-rx-widget: Add GAuth support

xerial:


airframe-rx/src/main/scala/wvlet/airframe/rx/RxOption.scala
```diff
@@ -54,6 +54,17 @@ trait RxOption[+A] extends Rx[Option[A]] {
     in.flatMap(f)
   }
 
+  def transformRxOption[B](f: Option[A] => RxOption[B]): RxOption[B] = {
```
xerial, line 57: This is not related to this PR, but I've found this transformation was missing.

xerial:
@shimamoto This is an attempt to extract common UI logic to airframe-rx-widget. I'll merge this PR in favor of post-review process, but let me know if you have any questions or ideas to improve it.

Issue #874, opened 2020-01-01
airframe-http-client: Add CircuitBreaker to http requests


xerial:
Add a fail-fast behavior to the http client to protect the target API service.

- This should be able to disable with noCircuitBreaker option.
- Support using a custom circuit breaker with withCircuitBreaker(...)

xerial:
This is done in #2190.

PR #1521, opened 2021-03-06
airframe-http: Add a default CircuitBreaker for js HTTP client

xerial:
This is for #874. Web browsers can be often disconnected from the server. Having a circuit breaker should be the default behavior to avoid accumulating too many error logs in the browser.


airframe-codec/src/main/scala/wvlet/airframe/codec/MessageCodecFactory.scala
```diff
@@ -17,7 +17,11 @@ case class MessageCodecFactory(codecFinder: MessageCodecFinder = Compat.messageC
 
   // Generate a codec that outputs objects as Map type. This should be enabled for generating JSON data
   def withMapOutput: MessageCodecFactory = {
-    this.copy(mapOutput = true)
+    if (mapOutput == true) {
+      this
```
xerial, line 21: Reduce the overhead of object creation if no change is required

xerial:
@shimamoto Could you review this change? Similar code is already used in our Conductor.

shimamoto (approved):
Looks good.

PR #1524, opened 2021-03-07
airframe-rx-html: Add beforeRender/beforeUnmount event handler

### LINKED ISSUE #1506: airframe-rx-html: Add event hooks before/after render
Something like https://vuejs.org/v2/api/#updated to start background RPC call process

xerial:
Resolves #1506

### DIFF
--- a/airframe-rx-html/src/main/scala/wvlet/airframe/rx/html/RxElement.scala
+++ b/airframe-rx-html/src/main/scala/wvlet/airframe/rx/html/RxElement.scala
@@ -26,6 +26,23 @@ abstract class RxElement(val modifiers: List[Seq[HtmlNode]] = List.empty) extend
     */
   def render: RxElement
 
+  /**
+    * Called right before rendering this RxElement begins.
+    *
+    * Override this method to define a custom event hook before rendering.
+    */
+  def beforeRender: Unit = {}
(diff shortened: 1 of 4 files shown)


airframe-rx-html/src/main/scala/wvlet/airframe/rx/html/RxElement.scala
```diff
@@ -34,10 +49,12 @@ abstract class RxElement(val modifiers: List[Seq[HtmlNode]] = List.empty) extend
     }
   }
 
-  def add(xs: HtmlNode*): RxElement =
+  def add(xs: HtmlNode*): RxElement = {
     new RxElement(xs :: modifiers) {
       override def render = self.render
```
takezoe, line 54: `beforeRender` and `beforeUnmount` also need to be overridden?

xerial: Nice catch. I'll fix it.

xerial: Actually, this part needed no fix if we change the logic of DOMRenderer to handle nested RxElements properly.

PR #1525, opened 2021-03-07
airframe-rx: Add Rx.cache

xerial:
This PR will add a handy method for caching the last event. 

This will simplify the code that needs to read RPC results periodically while reusing the last result when re-rendering the page. For example, we can write a page listing the remote worker nodes like this: 
```scala
private val nodeList = Rx.intervalMillis(1000).andThen(i => (rpc call to get a node list)).cache

def render = nodeList.map { list =>
   ... // If the cached data is available, it will be rendered quickly
}

// Rx.intervalMillis(1000) will stop properly when this DOM is unmounted
// via Rx operator update (e.g., Rx.variable.update ...)
```

@shimamoto Could your review this new operator? cc: @takezoe


shimamoto (approved):
Looks great! Thanks.

PR #1546, opened 2021-03-23
Path prefix versioning doesn't work in airframe-http

takezoe:
If airframe-http based web application has paths that are versioned by prefix and are nested as follows:

- `/v1/hello/world`
- `/v2/hello/world`

`RouteMatcher` yields an exception like this:

```
java.lang.IllegalArgumentException: Found multiple matching routes: /v1/hello/world, /v2/hello/world 
```

Looks like this check is affecting, but I'm not sure how we can fix it properly.
https://github.com/wvlet/airframe/blob/2c610f2acdfa50eb67a11631e630558dd193d81b/airframe-http/.jvm/src/main/scala/wvlet/airframe/http/router/RouteMatcher.scala#L70-L77

I think this issue is critical for airframe-http based applications because this means path prefix-based versioning is almost unavailable in airframe-http.


xerial:
This issue will be fixed with #1547

PR #1547, opened 2021-03-23
airframe-http: Fixes #1546 by adding path prefixes to disambiguate NFA nodes

xerial:
The problem was intermediate nodes were wrongly merged when building Automaton 
```
[OK]
v1 -> hello -> world (match)
v2 -> hello -> world (match) 

[Bug] 
v1 -> hello -> world (match)
v2 -/       \- world (match)
```

To solve this, added path prefixes to automaton nodes so that nodes belonging to different path prefixes will not be merged:
```
[With this PR]
v1 -> v1/hello -> v1/hello/world (match)
v2 -> v2/hello -> v2/hello/world (match)
```


takezoe (approved):
This works! Thank you!

PR #1565, opened 2021-03-30
Add missing args in airframe-codec docs

ka2kama:
- add missing args to  MessageCodec#toJson and MessageCodec#toMsgPack methods in doc.


xerial:
Merged. Thanks!

PR #1570, opened 2021-03-31
sbt-airframe: Generate ServiceJSClientRx for Scala.js + RPC + Rx

xerial:
With this PR, sbt-airframe will generate an RPC client (ServcieJSClientRx) for Scala.js that returns Rx[X] return type, instead of Future[X].

This client will simplify RPC calls from Scala.js like this:
```scala
val rpcClient = bind[ServiceJSClientRx]

rpcClient.ServiceApi.apiMethod(...).map { x =>
  div(s"result: ${x}")
}
```

Previously, we always needed to wrap Future with Rx:
```scala
val jsClient = bind[ServiceJSClient]

// We always need to import ExecutionContext for Scala.js
implicit val queue = scala.scalajs.concurrent.JSExecutionContext.queue

Rx.future(jsClient.ServiceApi.apiMethod(...)).map { x =>
  div(s"result: ${x}")
}
```

cc: @shimamoto @takezoe

### DIFF
--- a/airframe-http/.jvm/src/main/scala/wvlet/airframe/http/codegen/client/ScalaHttpClientGenerator.scala
+++ b/airframe-http/.jvm/src/main/scala/wvlet/airframe/http/codegen/client/ScalaHttpClientGenerator.scala
@@ -179,6 +179,7 @@ object ScalaJSClientGenerator extends HttpClientGenerator {
          |import wvlet.airframe.surface.Surface
          |import wvlet.airframe.http.js.JSHttpClient
          |import wvlet.airframe.http.HttpMessage.Request
+         |import wvlet.airframe.rx.RxStream
          |
          |${jsClientCls}
          |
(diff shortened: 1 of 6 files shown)


airframe-http/.jvm/src/main/scala/wvlet/airframe/http/codegen/client/ScalaHttpClientGenerator.scala
```diff
@@ -172,52 +172,95 @@ object ScalaJSClientGenerator extends HttpClientGenerator {
...
          |import scala.concurrent.Future
          |import wvlet.airframe.surface.Surface
          |import wvlet.airframe.http.js.JSHttpClient
          |import wvlet.airframe.http.HttpMessage.Request
+         |import wvlet.airframe.rx.RxStream
          |
-         |${cls}""".stripMargin
+         |${jsClientCls}
+         |
+         |${rpcClientCls}""".stripMargin
```
shimamoto, line 186: Both of them are created so we can choose, that's great.

shimamoto (approved):
Thank you

airframe-http/.jvm/src/main/scala/wvlet/airframe/http/codegen/client/ScalaHttpClientGenerator.scala
xerial: Right. When rendering DOM, Rx client is useful, but for calling RPC, which requires no DOM rendering, the existing ServiceJSClient is still necessary.

PR #1571, opened 2021-04-01
airframe-rx: Add Rx.startWith

xerial:
http://reactivex.io/documentation/operators/startwith.html

We can use this operator to render the loading state while waiting for RPC responses:
```scala
rpcCall.map( x => div(...))
.startWith(div("loading ..."))

// This is equivalent to
Rx.concat(Rx.single(div("loading ...")), rpcCall.map(x => div(...)) 
```


xerial:
cc: @shimamoto

PR #1572, opened 2021-04-02
airframe-rx: Add Rx.cache.getCurrent

xerial:
I've found it's necessary to implement components that have some public methods referencing Rx data:

```scala
class Selector extends RxElement { 
   private val list = getList.map(x => ...).cache 

   def selected = list.getCurrent.map(lst => lst(selectedIndex))
}
```


xerial:
cc: @shimamoto

Issue #1077, opened 2020-05-14
Milestone: Scala 3 (Dotty) Support


xerial:
# Scala 3 (dotty)
- [x] Use Scala 3.0.0 #1639 
- [x] Add a basic configuration
- [x] Extract AirSpec as an external project #1992 
- [x] Test Scala version sandwitch https://github.com/sbt/sbt/pull/5767
- [x] airframe-log support #867 #1381 
- [x] Surface support  #1384
   - [x] Surface.of[X] #1385 
   - [x] Surface.methodsOf[X] #1403 
- [x] airframe-di support #1649 
   - [x] Split source folder for Scala 2 and 3 #1405 
   - [x] Support Scala3 macros #1407 
   - [x] deprecate in-trait bindings #1653 
   - ~[ ] new A with DISupport https://github.com/wvlet/airframe/issues/1410~
   - ~[ ] Traverse Template node in Scala 3 for rewriting `new A with DISupport {}` code~ https://github.com/lampepfl/dotty/issues/10931
- [x] AirSpec #1659
- [x] Use TastyInspector
   - [x] Create ReflectSurface factory
   - [x] RuntimeSurface.of[X]
- [x] Run all AirSpec tests 
     - [x] parquet #2209 Need to resolve Surface of nested objects at runtime or pre-register Surfaces
     - [x] grpc (Need to register trait factories to start gRPC server. https://github.com/wvlet/airframe/issues/2194
       - [x] Scala 3.1.3-RC2 or later will be required for generating new class from a trait type T https://github.com/lampepfl/dotty/issues/11685) 
     - [x] airframe-log
     - [x] surface
        - [x] recursive surface (with type substitution)
        - [x] local surface inside object or class #1695 
        - [x] NamedParameterTest
        - [x] RecursiveHigherKindTypeTest #2052 
     - [x] di
     - [x] codec
     - [x] json
     - [x] msgpack
     - [x] canvas
     - [x] control
     - [x] ulid
     - [x] metrics
     - [x] rx
     - [x] okhttp (~Finagle server is used for testing~) #2440 #2865
     - [x] http-recorder (Finagle server is used internally) #1146 
- [x] airframe-config
- [x] airframe-http
- [x] airframe-http-router
- [x] airframe-http-grpc (codec is required)
  - [x] airframe-http-router
  - [x] airframe-http-codegen (Need to read Surfaces of RPC/Endpoint interfaces from classpath)
[...]

PR #1591, opened 2021-04-20
Add more Scala 3 support

xerial:
Add 7 more projects with Scala 3 support. 
- canvas
- control
- metrics
- msgpack
- json
- rx
- ulid

Some projects not in this list are using DI, Scala reflect, Scala2 macros. 
We need to support airframe-di for Scala 3 #1077


xerial:
cc: @wvlet/airframe-dev

Issue #1649, opened 2021-05-14
airframe-di: Next generation DI library


xerial:
- Support only constructor injection! 
  - If we use only constructor injection, no dependency to DI itself is necessary other than the design part
  - No need to worry about which injection to use (in-trait or constructor?). It will also reduce the learning cost. Half of the current documentation (https://wvlet.org/airframe/docs/airframe) will be unnecessary.
- Deprecate in-trait bindings
  - For supporting Scala 3 at ease. Writing macros for nested trait instantiation is the tricky part. If we use constructor injection, the code will be much simpler
- Migration plan
  - Retain the current airframe module as is.
  - Add a new module "airframe-di" so that the user can migrate to a new one at any timing. 
  - ~The package name will be `wvlet.airframe.di`~  => **Deprecate in-trait injection only from Scala 3!** Use the same package name to reduce the confusion.
  
- Usage stats of in-trait bindings (inside TD)
  - bind[X] a lot!!! 
  - bindLocal[X] only a few times
  - bindFactory[X] only a few times
  - So dropping bindLocal, bindFactory should be ok.

xerial:
Airframe DI for Scala 3 (without in-trait bindings) is released as 21.5.1

PR #1653, opened 2021-05-14
airframe-di: Next generation DI

xerial:
Based on idea #1649 

- [x] Deprecate in-trait binding
- [x] Add tests
- [x] Doc
- [ ] Scala 3 support


xerial:
Removing in-train injection to simplify the DI framework is done. I'll work on Scala3 support later. We may also need to refine some APIs (e.g. design.build { (A: App) => ... } might be better than design.build[A] { (A: App => ... }

PR #1686, opened 2021-05-22
Extract sbt-airframe as a separate GitHub project

xerial:
This PR extracts sbt-airframe module as a separate GitHub project https://github.com/wvlet/sbt-airframe. 

This is for simplifying the cross build settings. 
Currently, sbt-plugin must use only Scala 2.12, while the other airframe modules can be built with Scala 2.13 and Scala 3. 
Having a sbt-plugin inside the same project has caused a lot of dependency errors when switching to Scala 2.13 or Scala 3. 

With this change, we can make the default Scala version of Airframe at least to Scala 2.13. And also, cross building Scala.js projects with Scala 3 will be possible.


xerial:
cc: @wvlet/airframe-dev

PR #1703, opened 2021-05-26
airframe-http-recorder: Delete duplicated server creation methods

takezoe:


airframe-http-recorder/README.md
```diff
@@ -66,13 +66,15 @@ withResource(HttpRecorder.createReplayOnlyServer(recorderConfig)) { server =>
...
+You can record responses by code instead of recording real interactions with the server.
+
 ```scala
 import wvlet.airframe.http.recorder._
 import wvlet.airframe.control.Control._
 import com.twitter.finagle.http.{Request,Response}
 
 val recorderConfig = HttpRecorderConfig(sessionName="my-recording")
-val response = withResource(HttpRecorder.createProgrammableServer(recorderConfig)) { server =>
+val response = withResource(HttpRecorder.createReplayOnlyServer(recorderConfig)) { server =>
```
takezoe, line 77: Using ReplayOnlyServer as ProgrammableServer might be a little weird? InMemoryServer doesn't have variation of creation methods like recordOnly and replayOnly.

xerial: Ok, so how about renaming APIs as createRecordServer and createReplayServer?

xerial (approved):
Change itself is ok to merge.

airframe-http-recorder/README.md
takezoe: Sorry for late reply. That sounds good! I was also thinking that the following namings for consistency:

- createRecordServer
- createServer
- createInMemoryServer

What do you think?

xerial: I'm ok with these changes!

PR #1721, opened 2021-05-31
Make ULIDGenerator class public allowing Random injection

milanvdm:
We've noticed the ULID generation taking up to 10s due to threading issues on the default `SecureRandom` instance. 

When using a non-secure `Random` instance, it works fine. 

As far as I understand, as long as we simply use the generated ULIDs as an internal id that is only used for uniqueness and sorting, it is fine to use the non-secure `Random`

### DIFF
--- a/airframe-ulid/src/main/scala/wvlet/airframe/ulid/ULID.scala
+++ b/airframe-ulid/src/main/scala/wvlet/airframe/ulid/ULID.scala
@@ -184,7 +184,7 @@ object ULID {
     * @param timeSource a function that returns the current time in milliseconds (e.g. java.lang.System.currentTimeMillis())
     * @param random a function that returns a 80-bit random values in Array[Byte] (size:10)
     */
-  private class ULIDGenerator(random: () => Array[Byte]) {
+  class ULIDGenerator(random: () => Array[Byte]) {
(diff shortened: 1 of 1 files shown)


xerial:
Thanks! My benchmark in https://github.com/wvlet/airframe/tree/master/airframe-benchmark didn't use multiple threads, so I didn't notice this issue. 

If this use case is important, adding more utility methods like ULIDGenerator.newULID: ULID, newULIDString would be useful.

xerial:
@milanvdm I'll add ULIDGenerator.newULID, newULIDString to ULIDGeneartor for consistency with ULID methods.

milanvdm:
@xerial Thanks a bunch! :) 

How do releases work? Are there snapshot versions available or can a new minor version be released with these changes?

xerial:
We regularly release a new version at least once per month. 

A snapshot version is also available for every master branch commit: https://oss.sonatype.org/content/repositories/snapshots/org/wvlet/airframe/airframe-ulid_2.13/
The exact version name (e.g., 21.5.4-39-31c17913-SNAPSHOT) can be found in this log https://github.com/wvlet/airframe/runs/2712233322

I'm now adding more customization entries in #1722 so that you don't need to write your own generator.

PR #1759, opened 2021-07-09
Fix formatStacktrace dropping last line

### LINKED ISSUE #1754: airframe-log: formatStacktrace stripping off last line
LogFormatter.formatStacktrace() is stripping off the last line of the stacktrace.

The cause of this is:

```
        val filtered =
          stackTrace
            .split("\n") // Array
            .filter(stackTraceFilter)
            .sliding(2)
            .collect { case Array(a, b) if a != b => a }
```

Specifically, the sliding and collect. The collect only grabs the first of the 2 lines, so for the last pair, it grabs the 2nd last line and not the last.

This was causing an issue because we had an inner cause that only had a message and no stack trace, and the necessary message was getting stripped out.

renns:
Fixes #1754


xerial:
Thanks! Merged

PR #1760, opened 2021-07-12
airframe-http: Scala 3 support prep

xerial:


xerial:
For Scala 3, it hit an compilation error around HttpMessage https://github.com/lampepfl/dotty/issues/9270 

We need another approach to support Scala 3.

PR #1778, opened 2021-08-13
Upgrade to Finagle 21.8.0

xerial:
Use airframe-http-finagle client for proxy


airframe-http-recorder/src/main/scala/wvlet/airframe/http/recorder/HttpRecorder.scala
```diff
@@ -96,20 +93,11 @@ object HttpRecorder extends LogSupport {
...
 
-    (if (recorderConfig.destAddress.port == 443) {
-       clientBuilder.tls(SSLContext.getDefault)
-     } else {
-       clientBuilder
-     }).build()
+    Finagle.client
+      .withInitializer(_.withLabel("airframe-http-recorder-proxy"))
+      .newClient(recorderConfig.destAddress.hostAndPort)
+      .nativeClient
```
xerial, line 100: This change was necessary to properly configure SSL. This also avoids using the deprecated stack builder.

PR #1782, opened 2021-08-14
airframe-grpc: Support customizing CallOption for each RPC method

### LINKED ISSUE #1711: airframe-grpc: Support customizing io.grpc.CallOptions for each RPC call
For setting different credential for each RPC call

xerial:
Closes #1711


xerial:
SyncClient.withXXX can be used instead. We don't need this feature

PR #1813, opened 2021-08-30
airframe-arrow: Add ArrorFlight RPC interface

xerial:
ArrowFlight is using Protobuf for definign server/client. 
https://github.com/apache/arrow/blob/master/format/Flight.proto

This PR is an alternative interface of ArrowFlight without using any Protobuf. 
- FlightService server can be implemented in Scala using model classes. 
- FlightService client can be generated with sbt-airframe, which internally using MessagePack or JSON. If JSON is used, binary data need to be BASE64-encoded strings. 

Some learnings:
- ArrowFlights is using too many bytes array parameters. So the server/client developers still need some support for parsing byte array data.
- ArrowFlights is using client/server/bidi streaming a lot, but introducing pagenation might be better in terms of stability if it will be used for sending large volumes of data. 

discussion:
https://twitter.com/semanticbeeng/status/1432094625439887360

### DIFF
--- a/airframe-arrow/src/main/scala/wvlet/airframe/arrow/flight/api/FlightService.scala
+++ b/airframe-arrow/src/main/scala/wvlet/airframe/arrow/flight/api/FlightService.scala
@@ -1,0 +1,322 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package wvlet.airframe.arrow.flight.api
+
+import wvlet.airframe.http.RPC
+import wvlet.airframe.rx.RxStream
+
+/*
+ * A flight service is an endpoint for retrieving or storing Arrow data. A
+ * flight service can expose one or more predefined endpoints that can be
+ * accessed using the Arrow Flight Protocol. Additionally, a flight service
+ * can expose a set of actions that are available.
+ */
+@RPC
+trait FlightService {
+  import FlightService._
+
+  /*
+   * Handshake between client and server. Depending on the server, the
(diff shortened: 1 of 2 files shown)


SemanticBeeng:
> * server/client developers still need some support for parsing byte array data.

`Apache Arrow` is also implemented in JavaScript. https://arrow.apache.org/docs/js/
Could this be used in the generated bindings?

SemanticBeeng:
> * ArrowFlights is using client/server/bidi streaming a lot, but introducing pagenation might be better in terms of stability if it will be used for sending large volumes of data.

`Apache Arrow` has this `RecordBatch`: https://gist.github.com/SemanticBeeng/0e9f9ed40866a5f844be13ab6ef3469e
I have the sense the `Arrow Flight` is using it behind the `gRPC` streaming.

Wondering how this could fit with `airframe`.

xerial:
@SemanticBeeng If gRPC streaming is not used, sending data from server to browses can be supported if we launch Finagle (HTTP/1) backend. sbt-airframe already can generate JS client for RPC.

For doing so, we need to have different API, which doesn't require streaming (e.g., pagination-only)

SemanticBeeng:
> If gRPC streaming is not used, sending data from server to browses can be supported if

So are you saying that `airframe-rpc` cannot support this https://github.com/wvlet/airframe/pull/1813#issuecomment-908722225 in `gRPC` binary streaming mode ?

I am somewhat aware of the limitations of `gRPC` in browsers but was left with the impression that this project is somehow improving on that. :thinking: 

If `gRPC` binary streaming mode cannot be supported with native `apache arrow` then indeed we can use a pagination based api using any of the backend frameworks.

Please advise @xerial

xerial:
For using gRPC backend of grpc-java, we can implement Arrow-Flight like service in this PR using Airframe RPC, which supports all streaming modes of gRPC.

As there seems no active plan in gRPC and web browsers to support client-side streaming (e.g., Sending request parameters from a web browser to a server asynchronously), the current Arrow-flight RPC design cannot be used for supporting JavaScripts running at web browsers. Arrow Flight should have a version that has no client-side streaming and bidirectional streaming if we need web browser support.

xerial:
gRPC and grpc-java are not suited to sending large volumes of byte data. For example, grpc-java uses just 4k for the gRPC message size. We can increase the limit, but sending large binary data is apparently not the central use cases of gRPC. 

I think the Apach Arrow protocol itself needs to be redesigned (e.g., creating non gRPC endpoint for sending binary data). I'll close this PR as supporting Apache Arrow with gRPC is not so promising at this moment.

PR #2038, opened 2022-01-28
Bump scalafmt and fix format

exoego:
Supersedes #2026

### DIFF
--- a/airframe-http-finagle/src/main/scala/wvlet/airframe/http/finagle/FinagleServer.scala
+++ b/airframe-http-finagle/src/main/scala/wvlet/airframe/http/finagle/FinagleServer.scala
@@ -299,9 +299,8 @@ trait FinagleServerFactory extends AutoCloseable with LogSupport {
 
   /**
     * Override this method to customize Finagle Server configuration.
-    * @deprecated(description
-    * = "Use FinagleServerConfig.withServerInitializer", since = "19.11.0")
     */
+  @deprecated(message = "Use FinagleServerConfig.withServerInitializer", since = "19.11.0")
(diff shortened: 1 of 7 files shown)


exoego:
@xerial 
I checked https://docs.scala-lang.org/overviews/scaladoc/for-library-authors.html and https://en.wikipedia.org/wiki/Help:Cheatsheet

Scaladoc uses Wiki-stye markup, not a Markdown.
So I assume scalafmt escaped `=` into `\=`, where `=` is not for heading `=Heading=`.
`|` is also have special meaning in Wiki markup, so escaped.

Instead of `\` escape, I changed single line code into monospace:
```
`single line`
```
and multiple-lines code into code block:

```
{{{
val a = 1
val b = 1
val c = 1
}}}
```

airframe-http-finagle/src/main/scala/wvlet/airframe/http/finagle/FinagleServer.scala
```diff
@@ -299,9 +299,8 @@ trait FinagleServerFactory extends AutoCloseable with LogSupport {
 
   /**
     * Override this method to customize Finagle Server configuration.
-    * @deprecated(description
-    * = "Use FinagleServerConfig.withServerInitializer", since = "19.11.0")
     */
+  @deprecated(message = "Use FinagleServerConfig.withServerInitializer", since = "19.11.0")
```
exoego, line 303: I assume `@deprecated` were accidentally in Scaladoc, so extracted it.

xerial:
Thanks! I still don’t remember correct scalacon syntax 😅

PR #2043, opened 2022-01-30
Share same scalafmt config between airframe and airspec

exoego:
Supersedes #2042 

I think this makes easy to version up scalafmt


xerial:
Merged. Thanks!

PR #2049, opened 2022-02-02
Use munit temporarily to avoid airspec <-> surface circular reference

exoego:
When working on Scala 3 support in `surface` module, I found `DOTTY=1 sbt surfaceJVM/test` raises errors like:
<details>
<pre><codE>
[error] Test suite wvlet.airframe.surface.ZeroTest failed with java.lang.NoClassDefFoundError: Could not initialize class wvlet.airframe.metrics.ElapsedTime$.
[error] This may be due to the ClassLoaderLayeringStrategy (ScalaLibrary) used by your task.
[error] To improve performance and reduce memory, sbt attempts to cache the class loaders used to load the project dependencies.
[error] The project class files are loaded in a separate class loader that is created for each test run.
[error] The test class loader accesses the project dependency classes using the cached project dependency classloader.
[error] With this approach, class loading may fail under the following conditions:
[error] 
[error]  * Dependencies use reflection to access classes in your project's classpath.
[error]    Java serialization/deserialization may cause this.
[error]  * An open package is accessed across layers. If the project's classes access or extend
[error]    jvm package private classes defined in a project dependency, it may cause an IllegalAccessError
[error]    because the jvm enforces package private at the classloader level.
[error] 
[error] These issues, along with others that were not enumerated above, may be resolved by changing the class loader layering strategy.
[error] The Flat and ScalaLibrary strategies bundle the full project classpath in the same class loader.
[error] To use one of these strategies, set the  ClassLoaderLayeringStrategy key
[error] in your configuration, for example:
[error] 
[error] set surfaceJVM / Test / classLoaderLayeringStrategy := ClassLoaderLayeringStrategy.ScalaLibrary
[error] set surfaceJVM / Test / classLoaderLayeringStrategy := ClassLoaderLayeringStrategy.Flat
[error] 
[error] See ClassLoaderLayeringStrategy.scala for the full list of options.
</code></pre>
</details>

I assume this is because 
1. `surface` is currently tested with `airspec`
2. `airspec` depends on `surface` (circular reference)
3. Adding missing classes for Scala 3, already used in `airspec`, brakes `airspec`

This error makes it difficult to work on Scala 3 support.
So I propose temporal usage of [munit](https://github.com/scalameta/munit) for testing `surface`.
When `surface` get Scala 3 support, go back to `airspec`.

This PR is tested with `DOTTY=1 sbt surfaceJVM/test`.

### DIFF
--- a/airframe-surface/.jvm/src/test/scala-3/wvlet/airframe/surface/reflect/TastySurfaceFactoryTest.scala
+++ b/airframe-surface/.jvm/src/test/scala-3/wvlet/airframe/surface/reflect/TastySurfaceFactoryTest.scala
@@ -1,12 +1,13 @@
 package wvlet.airframe.surface.reflect
 
-import wvlet.airspec.AirSpec
+import wvlet.log.LogSupport
+import wvlet.airframe.surface.AirSpecBridge
 
 case class Person(id: Int, name: String) {
   def hello: String = "hello"
 }
 
-object TastySurfaceFactoryTest extends AirSpec {
+class TastySurfaceFactoryTest extends munit.FunSuite with AirSpecBridge with LogSupport {
 
   test("of[A]") {
     val s = TastySurfaceFactory.of[Person]
--- a/airframe-surface/.jvm/src/test/scala/wvlet/airframe/surface/AbstractTypeTest.scala
+++ b/airframe-surface/.jvm/src/test/scala/wvlet/airframe/surface/AbstractTypeTest.scala
@@ -1,8 +1,6 @@
 package wvlet.airframe.surface
 
-import wvlet.airspec.AirSpec
-
-object AbstractTypeTest extends AirSpec {
+class AbstractTypeTest extends munit.FunSuite {
 
   trait Abst {
     def hello = "hello abst"
@@ -13,9 +11,9 @@ object AbstractTypeTest extends AirSpec {
 
   test("object factory of an abstract type impl") {
     val s = Surface.of[AbstImpl]
-    s.objectFactory shouldBe defined
+    assert(s.objectFactory.isDefined)
 
     val a = s.objectFactory.get.newInstance(Seq.empty).asInstanceOf[Abst]
-    a.hello shouldBe "hello impl"
+    assertEquals(a.hello, "hello impl")
   }
 }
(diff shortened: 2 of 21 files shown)


airframe-surface/.jvm/src/test/scala-3/wvlet/airframe/surface/reflect/TastySurfaceFactoryTest.scala
```diff
@@ -1,12 +1,13 @@
...
-import wvlet.airspec.AirSpec
+import wvlet.log.LogSupport
+import wvlet.airframe.surface.AirSpecBridge
 
 case class Person(id: Int, name: String) {
   def hello: String = "hello"
 }
 
-object TastySurfaceFactoryTest extends AirSpec {
+class TastySurfaceFactoryTest extends munit.FunSuite with AirSpecBridge with LogSupport {
```
exoego, line 10: Note) `MUnit` requires all test cases must be `class`

airframe-surface/.jvm/src/test/scala/wvlet/airframe/surface/AbstractTypeTest.scala
```diff
@@ -13,9 +11,9 @@ object AbstractTypeTest extends AirSpec {
 
   test("object factory of an abstract type impl") {
     val s = Surface.of[AbstImpl]
-    s.objectFactory shouldBe defined
+    assert(s.objectFactory.isDefined)
```
exoego, line 14: Note) `MUnit` has only JUnit-like assertions.

build.sbt
```diff
@@ -436,6 +436,8 @@ lazy val surface =
     .settings(
       name        := "airframe-surface",
       description := "A library for extracting object structure surface",
+      libraryDependencies -= "org.wvlet.airframe" %%% "airspec" % AIRSPEC_VERSION % Test,
+      libraryDependencies += "org.scalameta" %%% "munit" % "0.7.29" % Test,
```
xerial, line 440:
```suggestion
      // TODO: This is a temporaly solution. Use AirSpec after Scala 3 support of Surface is completed
      libraryDependencies += "org.scalameta" %%% "munit" % "0.7.29" % Test,
```

xerial:
ok. Undrestood. Added a comment to clarify why we are using munit.

PR #2056, opened 2022-02-04
[build] Make it easier on cross-build while working on Scala 3 support

exoego:
When working on Scala 3 support, I often need to run tests over Scala 3 and 2, like:
```
$ DOTTY=1 sbt 
sbt> ++surfaceJVM/test
```
to confirm behaviours both on Scala 3 and 2.
However, `++COMMAND` is failing on Scala 2 right now, because settings are only configured for Scala 3 due to hard-coded `if (DOTTY)`.

This PR replaces `DOTTY` criteria as much as possible to `scalaVersion.value.startsWith("3.")` so cross-build can be

<del>And, `dottyTest/test` now runs all tests on dotty-support-ongoing project.</del>
`projectDotty` already exists for that purpose.

### DIFF
--- a/airspec/build.sbt
+++ b/airspec/build.sbt
@@ -83,7 +83,7 @@ val buildSettings = Seq[Setting[_]](
     "-feature",
     "-deprecation"
   ) ++ {
-    if (DOTTY) {
+    if (scalaVersion.value.startsWith("3.")) {
       Seq.empty
     } else {
       Seq(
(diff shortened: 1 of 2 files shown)


xerial:
We also need to fix some release scripts like https://github.com/wvlet/airframe/blob/master/.github/workflows/release.yml#L32

exoego:
IIUC, I think no need to fix `+ projectJVM/publishSigned` because `if (DOTTY)` remains for 

- `crossScalaVerions` 
- `scalaVersion`

so the below command are not affected by this PR. 

```
./sbt "; + projectJVM/publishSigned;" # publishes for Scala 2.12 and 2.13
DOTTY=true ./sbt projectDotty/publishSigned # publishes for Scala 3
```

xerial:
Ah. OK. This PR still preserve DOTTY variable and crossScalaVersions setting.

xerial:
It's good to merge. Thanks

PR #2079, opened 2022-02-18
Support positive base duration in TimeWindow

takezoe:


xerial (approved):
Ah. Good catch. Thanks

PR #2087, opened 2022-02-23
Slow down awssdk update

exoego:
Scala Steward now can configure update frequency granularly.
https://github.com/scala-steward-org/scala-steward/pull/2515


xerial:
Great! I think we can apply the same config to scala-ulid (from Chatwork) as well.

PR #2159, opened 2022-05-06
AirSpec: Add async test support

xerial:
Added support for running AirSpec tests returning Future values. 

- [x] Refactor AirSpecTaskRunner
- [x] Return Future from AirSpecTaskRunner
- [x] concatenate to continuation in Scala.js
- [x] Manage Future objects generated at nested tests


xerial:
With this feature, writing async UI tests or RPC tests becomes possible. 
cc: @exoego @takezoe @shimamoto

PR #2162, opened 2022-05-07
Adopt to asynchronous behavior of AirSpec tests

xerial:


xerial:
This is a follow-up work for #2159 so that tests will not fail in the next AirSpec version

PR #2191, opened 2022-05-27
Drop Java8 support and use Java11 http client by default

xerial:
As we no longer have projects that require Java8, I'd like to drop Java8 support for ease of maintenance. And also, dropping Java8 support enables us to use the standard Java11 HttpClient (#2188), which can be used without dependency like Finagle/OkHttp. 

Note: airframe-http-finagle/okhttp can be used without any change. There is no plan to drop the HTTP clients in these modules.

cc: @wvlet/airframe-dev @exoego

### DIFF
--- a/airframe-http/src/main/scala/wvlet/airframe/http/Http.scala
+++ b/airframe-http/src/main/scala/wvlet/airframe/http/Http.scala
@@ -15,6 +15,7 @@ package wvlet.airframe.http
 
 import wvlet.airframe.http.HttpBackend.DefaultBackend
 import wvlet.airframe.http.HttpMessage.{Request, Response}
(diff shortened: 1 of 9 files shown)


airframe-http/src/main/scala/wvlet/airframe/http/Http.scala
```diff
@@ -44,12 +45,6 @@ object Http {
     */
   def client: HttpClientConfig = HttpClientConfig()
 
-  /**
-    * Create the default HTTP sync client for the target server address
-    */
-  def clientFor(serverAddress: String): wvlet.airframe.http.client.SyncClient =
```
xerial, line 50: Removing because it is a duplicate of Http.client.newSyncClient(serverAddress)

PR #2211, opened 2022-06-01
airframe-parquet: Scala 3 support

xerial:


xerial:
- AnyCodec now accepts knownSurfaces parameter to resolve Surface from Class[_] at runtime
- Add Resource.use for using loan pattern
- Parquet.newRecordWriter accepts knownSurface parameter to support nested objects

PR #2260, opened 2022-06-20
Avoid modifying surface cache while updating it

som-snytt:
Community build https://github.com/scala/community-build/pull/1568 noticed that delegating to `ConcurrentHashMap` in https://github.com/scala/scala/pull/10027 makes `ReflectSurfaceFactory.apply` fail due to recursive update of cache.

This commit refactors so that the "update" function returns the surface or null for the LazySurface case. (Where null means nothing to cache.)

I had trouble testing: first to get the build to use my local Scala, which requires setting scalaHome in the subproject settings, then with getting the `test` run to use my fix in the code base, which I did not resolve. So I have not seen the test go green; the stack trace is the old code. Does cross compilation somehow pick up the old library? I did not investigate further at this point.

### DIFF
--- a/airframe-surface/.jvm/src/main/scala-2/wvlet/airframe/surface/reflect/ReflectSurfaceFactory.scala
+++ b/airframe-surface/.jvm/src/main/scala-2/wvlet/airframe/surface/reflect/ReflectSurfaceFactory.scala
@@ -36,6 +36,10 @@ object ReflectSurfaceFactory extends LogSupport {
     new ConcurrentHashMap[TypeName, Seq[MethodSurface]].asScala
   private[surface] val typeMap = new ConcurrentHashMap[Surface, ru.Type].asScala
 
+  //private def surfaceCacheCheat = surfaceCache.asInstanceOf[scala.collection.convert.JavaCollectionWrappers.JConcurrentMapWrapper]
+  import language.reflectiveCalls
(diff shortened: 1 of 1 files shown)


xerial:
@som-snytt Thanks for the pointer to the change of ConcurrentHashMap. I guess it will not be back-ported to Scala 2.12, and will be available in Scala 2.13.10.

A reason you see an old code reference is that AirSpec, a testing library used for testing Airframe, embeds the source code of Surface.

Anyway, the problem is that Surface needs to avoid recursive updates of cache while running getOrElseUpdate. It will be a big change, though.

xerial:
@som-snytt Let me try a fundamental fix this week by changing the way of caching Surfaces.

som-snytt:
OK. My other idea is to inline `surfaceCache.underlying.computeIfAbsent`. (cast to the wrapper required)  I'll push that if it works.

Actually, the cast is hindered by
```
// not private[convert] because `WeakHashMap` uses JMapWrapper
private[collection] object JavaCollectionWrappers extends Serializable {
```
It should not be private at all, for just this use case! Or all wrappers should extend an interface that exposes the underlying collection.

xerial:
Applied a potential fix in #2261. I haven't checked this build using a snapshot version of Scala 2.13. Need to wait for the next community build run.

som-snytt:
@xerial Thanks!

PR #2284, opened 2022-06-28
Upgrade to Scala 3.1.3

xerial:
- [ ] Fix wwo test failures https://github.com/wvlet/airframe/pull/2284/checks?check_run_id=7094398971


xerial:
After upgrading Airspec, the error is resolved. It's a bit weird, but anyway I'll merge this PR. If any error occurs, we need to fix it later.

PR #2308, opened 2022-07-08
Use Scala 2.12.16 for testing

xerial:


xerial:
Scala 3 test is still unstable ...

PR #2395, opened 2022-09-05
airframe-surface: Scala3 higher-kind type support

xerial:
- Remove `import dotty.tools.dotc.core.{Types as DottyTypes}`, which is available only in JVM, for supporting Scala.js
- Use pattern matching with ParamRef and TypeBounds instead of asInstanceOf class check


xerial:
@exoego After merging this PR, a remaining fix of airframe-surface for supporting Scala.js in Scala 3 is only #2396

PR #2487, opened 2022-10-18
Trigger release workflows when publishing a release note

xerial:


xerial:
It seems using script/release.rb is better for pushing tags instead of having multiple steps.

PR #2494, opened 2022-10-20
Allow to specify JMX MBean property name explicitly

takezoe:

PR #2506, opened 2022-10-22
Update airframe-jmx doc to describe name attribute of JMX annotation

takezoe:
Follow up to https://github.com/wvlet/airframe/pull/2494


xerial:
Thanks!

FYI. If the PR touches *.md file, release-drafter automatically add `doc` label to the PR https://github.com/wvlet/airframe/blob/master/.github/release-drafter.yml#L36-L39

PR #2507, opened 2022-10-22
Remove the generated modJVM and modJS definitions

lolgab:
sbt-crossproject now generates the modJVM and modJS lazy vals so can remove them.
This helps lowering the amount of code needed to support Scala Native


exoego (approved):
LGTM

PR #2521, opened 2022-10-31
airframe-sql: Fixes table/relation alias handling in TypeResolver 

xerial:
- Added `qualifier` parameter to ResolvedAttribute for tracking aliases
- AliasedRelation.output attributes returns a list of ResolvedAttributes with a qualifier (i.e. aliases)
- Fixed column name match when the context database is in the qualified name (QName)


yajirobee (approved):
nice 👍

PR #2533, opened 2022-11-05
airframe-http: Add unified http logging interface

xerial:
- Use the same logging logic between server and clients
- Use the same logging logic for the server-side backends, netty, finagle, grpc
- Support logging customization (exclude headers)


xerial:
superseded by #2912

PR #2537, opened 2022-11-10
Migrate to Docusaurus 2

xerial:


xerial:
This turned out to be a big effort. The migration needs to be finished at once. Wil try it again later

PR #2539, opened 2022-11-11
Support PrefixedULID case class in ULID

### LINKED ISSUE #2504: ULID with prefixes
ULID with prefix (e.g., P:01AN4Z07BY79KA1307SR9X4MV3, (resource type):(ULID)) format is convenient to distinguish resource types while maintaining timestamp orders. 

For example: 
```scala
case class ResourceID(prefix:String, ulid: ULID) 
```

yuokada:
Close #2504

### DIFF
--- a/airframe-ulid/src/main/scala/wvlet/airframe/ulid/ULID.scala
+++ b/airframe-ulid/src/main/scala/wvlet/airframe/ulid/ULID.scala
@@ -330,6 +330,6 @@ object ULID {
 
 }
 
-case class ResourceID(prefix: String, ulid: ULID, delimiter: Char = ':') {
-  override def toString: String = s"$prefix$delimiter$ulid"
+case class PrefixedULID(prefix: String, ulid: ULID, delimiter: Char = ':') {
+  override def toString: String = s"${prefix}${delimiter}${ulid}"
 }
--- a/airframe-ulid/src/main/scala/wvlet/airframe/ulid/ULID.scala
+++ b/airframe-ulid/src/main/scala/wvlet/airframe/ulid/ULID.scala
@@ -329,7 +329,3 @@ object ULID {
   }
 
 }
-
-case class PrefixedULID(prefix: String, ulid: ULID, delimiter: Char = ':') {
-  override def toString: String = s"${prefix}${delimiter}${ulid}"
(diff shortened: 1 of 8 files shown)


airframe-ulid/src/main/scala/wvlet/airframe/ulid/ULID.scala
```diff
@@ -329,3 +329,7 @@ object ULID {
   }
 
 }
+
+case class ResourceID(prefix: String, ulid: ULID, delimiter: Char = ':') {
+  override def toString: String = s"$prefix$delimiter$ulid"
```
xerial, line 334: Wrap individual variables with `${...}` for consistency with other code. Actually, always wrapping it is preferred for readability

airframe-ulid/src/main/scala/wvlet/airframe/ulid/ULID.scala
```diff
@@ -329,3 +329,7 @@ object ULID {
   }
 
 }
+
+case class ResourceID(prefix: String, ulid: ULID, delimiter: Char = ':') {
```
xerial, line 333: ResourceID is a too specific name. As a library, having a generic name is preferred. 

- PrefixedULID 
- PULID 
- etc. 

What do you think?

yuokada: I prefer PrefixedULID. In my opinion,  it is difficult to quickly understand what `PULID` is.

yuokada:
@xerial Fixed the code. PTAL.

xerial:
Thanks. Create a separate class for PrefixedULID.

A general use case is using PrefixedULID as a case class parameter (in airframe-http), so in addition, we need to:
- Add a codec for PrefixedULID to serialize it as a string (See ULIDCodec as an example)
- Register the codec for PrefixedULID in the following code:
image

xerial:
Discussion note
- Only support `:` as the delimiter
- Take the last `:` when unpack

xerial:
@yuokada Added more test code and utility methods. PTAL

yuokada:
@xerial Checked. Thanks for improving my code.

PR #2540, opened 2022-11-11
Support Customer Exclude Headers when creating AccessLogFilter

yuokada:


xerial:
Sorry. A large refactoring of logging is on-going at #2533

yuokada:
Ok, understand. I will close this. If the refactoring is finished, I will send another one if necessary.

PR #2555, opened 2022-11-17
airframe-sql: Support select .. having without group by

xerial:


xerial:
@takezoe I didn't know SQL can support SELECT ... HAVING  without setting any GROUP BY keys :)

PR #2556, opened 2022-11-17
Include inputAttributes of all relations in UNION

takezoe:
Include inputAttributes of all relations in UNION's inputAttributes.


xerial:
Thanks!

PR #2565, opened 2022-11-19
openapi: Fixes #2562. Find method default arguments only when RPC interface is a class

### LINKED ISSUE #2562: openapi: Fix required parameter labeling
If a default parameter is provided for an API request param, it should be optional. Currently, required label is used.

xerial:
- openapi: Fixes #2562. Find method default arguments only when RPC interface is a class
- wip


xerial:
@suprithIUB This will fix duplicate OpenAPI parameter generation for the same object.

PR #2578, opened 2022-11-23
airframe-sql: Allow ResolvedAttribute to have multiple source columns

takezoe:
for handling UNION (and maybe further cases)

### DIFF
--- a/airframe-sql/src/main/scala/wvlet/airframe/sql/model/ResolvedPlan.scala
+++ b/airframe-sql/src/main/scala/wvlet/airframe/sql/model/ResolvedPlan.scala
@@ -52,7 +52,9 @@ case class TableScan(table: Catalog.Table, columns: Seq[Catalog.TableColumn], no
 
 case class Alias(name: String, resolvedAttribute: ResolvedAttribute)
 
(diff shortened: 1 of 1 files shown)


airframe-sql/src/main/scala/wvlet/airframe/sql/model/ResolvedPlan.scala
```diff
@@ -53,46 +52,53 @@ case class TableScan(table: Catalog.Table, columns: Seq[Catalog.TableColumn], no
...
-        s"${q}.${name}:${dataType} <- ${t.name}.${c.name}"
-      case (None, Some(t), Some(c)) if c.name == name =>
-        s"${t.name}.${name}:${dataType}"
-      case (None, Some(t), Some(c)) =>
-        s"${name}:${dataType} <- ${t.name}.${c.name}"
+    (qualifier, sourceColumns) match {
+      case (Some(q), columns) if columns.nonEmpty =>
+        columns
+          .map { c =>
+            s"${c.table.name}.${c.column.name}"
```
xerial, line 92: Ok. I'll add SourceColumn.fullName for this

PR #2587, opened 2022-11-24
airframe-sql: Fix duplicated join keys handling

takezoe:
ouptputAttribute stays unresolved if the column is a duplicated key in equi join.

```scala
val p = analyze("select B.id from A inner join B on A.id = B.id")

println(p.outputAttributes) // => List(SingleColumn(UnresolvedAttribute(B.id)))
```

With this fix, outputAttributes will be `List(id:long <- B.id)`.


xerial:
Merged. Thanks

PR #2589, opened 2022-11-25
airframe-sql: Intersect should have attributes of all source relations

takezoe:
In the same way as Union

### DIFF
--- a/airframe-sql/src/main/scala/wvlet/airframe/sql/model/LogicalPlan.scala
+++ b/airframe-sql/src/main/scala/wvlet/airframe/sql/model/LogicalPlan.scala
@@ -623,7 +623,6 @@ object LogicalPlan {
           )
         }
       }
-      println("** out: " + out)
       out
     }
   }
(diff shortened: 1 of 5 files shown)


xerial:
Build is failing for Scala 3 🤔

xerial:
Removed debug logs

takezoe:
Ah, remaining debug code was not Scala3 compatible. 🙇‍♂️

airframe-sql/src/main/scala/wvlet/airframe/sql/model/LogicalPlan.scala
```diff
@@ -601,15 +601,30 @@ object LogicalPlan {
   sealed trait SetOperation extends Relation {
     override def children: Seq[Relation]
   }
-  case class Intersect(relations: Seq[Relation], nodeLocation: Option[NodeLocation]) extends SetOperation {
+  case class Intersect(
+      relations: Seq[Relation],
+      resolvedOutputs: Option[Seq[Attribute]],
```
xerial, line 606: Thanks!

PR #2591, opened 2022-11-27
airframe-http: Add RxFilter interface

xerial:


airframe-http-finagle/src/main/scala/wvlet/airframe/http/finagle/package.scala
```diff
@@ -204,4 +207,21 @@ package object finagle {
...
     resp
   }
+
+  /**
+    * Convert twitter Future into Rx
+    * @param twitterFuture
+    * @tparam A
+    * @return
+    */
+  def convertToRx[A](twitterFuture: Future[A]): RxStream[A] = {
```
xerial, line 217: This should move to an independent object?

airframe-http/src/main/scala/wvlet/airframe/http/RxFilter.scala
```diff
@@ -0,0 +1,103 @@
...
+import wvlet.airframe.http.HttpMessage.{Request, Response}
+import wvlet.airframe.rx.Rx
+
+import scala.util.control.NonFatal
+
+/**
+  * [[RxContext]] is a service interface for processing request and returning `Rx[Response]`.
+  */
+trait RxContext {
+  private[http] def backend: HttpBackend[Request, Response, Rx]
```
xerial, line 25: The current HttpBackend implementation is a bit heavy. We only need a request/response type converter and thread-local storage. It would be better to introduce a lightweight interface.

PR #2593, opened 2022-11-28
airframe-codec: Closes #2437 Use key-value map output by default

### LINKED ISSUE #2437: airframe-codec: Use json object structure mapping by default
The current default behavior of MessageCodec is using array output of object fields, but it should be map output unless the output size is the concern of the user.

xerial:


xerial:
@takezoe @yuokada Given a case class `A(p1, p2, ..)`, producing array output `[p1, p2, ..]` in MessageCodec is not robust in terms of object parameter change and object-json mappings. This PR changes the default output shape to key-value maps like `{p1 -> ..., p2 -> ... }`.

PR #2597, opened 2022-11-29
airframe-sql: Prevent resolving Identifiers of column alias

takezoe:
Prevent resolving `Identifier`s of alias by making them `String`.

### DIFF
--- a/airframe-sql/src/main/scala/wvlet/airframe/sql/analyzer/TypeResolver.scala
+++ b/airframe-sql/src/main/scala/wvlet/airframe/sql/analyzer/TypeResolver.scala
@@ -236,7 +236,7 @@ object TypeResolver extends LogSupport {
             resolvedColumns += r
           case r: ResolvedAttribute if alias.nonEmpty =>
             resolvedColumns += ResolvedAttribute(
-              alias.get.sqlExpr,
+              alias.get,
               r.dataType,
               r.qualifier,
               r.sourceColumns,
@@ -256,8 +256,8 @@ object TypeResolver extends LogSupport {
     attribute match {
       case SingleColumn(r: ResolvedAttribute, None, _, _) =>
         r
-      case SingleColumn(r: ResolvedAttribute, Some(alias: Identifier), _, _) =>
-        r.withAlias(alias.value)
+      case SingleColumn(r: ResolvedAttribute, Some(alias), _, _) =>
+        r.withAlias(alias)
       case other => other
     }
   }
@@ -273,8 +273,8 @@ object TypeResolver extends LogSupport {
       expr: Expression,
       inputAttributes: Seq[Attribute]
   ): List[Expression] = {
-    trace(s"findMatchInInputAttributes: ${expr}, inputAttributes: ${inputAttributes}")
-    def lookup(name: String): List[Attribute] = {
+    info(s"findMatchInInputAttributes: ${expr}, inputAttributes: ${inputAttributes}")
+    def lookup(name: String): List[Expression] = {
       QName(name, None) match {
         case QName(Seq(db, t1, c1), _) if context.database == db =>
           inputAttributes.collect {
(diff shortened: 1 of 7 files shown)


airframe-sql/src/main/scala/wvlet/airframe/sql/model/Expression.scala
```diff
@@ -239,12 +239,12 @@ object Expression {
   }
   case class SingleColumn(
       expr: Expression,
-      alias: Option[Expression],
+      alias: Option[String],
```
xerial, line 242: Make sense. A concern is handling various quotation types (e.g., unquoted/quoted identifier) in SQLGenerator.

airframe-sql/src/main/scala/wvlet/airframe/sql/model/LogicalPlan.scala
```diff
@@ -539,7 +539,7 @@ object LogicalPlan {
           query.outputAttributes.zip(aliases).map { case (in, alias) =>
             SingleColumn(
               in,
-              Some(alias),
+              Some(alias.sqlExpr),
```
xerial, line 542: If the alias is Identifier, using `Identifier.value` would be better to remove double quotation. 

`Identifier.value` method for case class may not work in Scala 2.13/3, so renaming value method to such as `.idString` might be necessary.

airframe-sql/src/main/scala/wvlet/airframe/sql/parser/SQLGenerator.scala
```diff
@@ -364,7 +364,7 @@ object SQLGenerator extends LogSupport {
       case SingleColumn(ex, alias, _, _) =>
         val col = printExpression(ex)
         alias
-          .map(x => s"${col} AS ${printExpression(x)}")
+          .map(x => s"${col} AS ${x}")
```
xerial, line 367: If `x` contains special characters or reserved words, it should be double-quoted. 

https://github.com/wvlet/airframe/blob/d3b1fbcd386001b008c710b41dd59a2bbb5cbbf9/airframe-sql/src/main/antlr4/wvlet/airframe/sql/parser/SqlBase.g4#L438

airframe-sql/src/main/scala/wvlet/airframe/sql/model/Expression.scala
xerial: For clarity, let's mention that alias can be unquoted/quoted strings.

airframe-sql/src/main/scala/wvlet/airframe/sql/parser/SQLGenerator.scala
xerial: Never mind. x (.sqlExpr) is already double quoted

airframe-sql/src/main/scala/wvlet/airframe/sql/model/Expression.scala
xerial, line 242:
```suggestion
      // quoted or unquoted string 
      alias: Option[String],
```

airframe-sql/src/main/scala/wvlet/airframe/sql/analyzer/TypeResolver.scala
```diff
@@ -273,8 +273,8 @@ object TypeResolver extends LogSupport {
       expr: Expression,
       inputAttributes: Seq[Attribute]
   ): List[Expression] = {
-    trace(s"findMatchInInputAttributes: ${expr}, inputAttributes: ${inputAttributes}")
-    def lookup(name: String): List[Attribute] = {
+    info(s"findMatchInInputAttributes: ${expr}, inputAttributes: ${inputAttributes}")
```
xerial, line 276:
```suggestion
    debug(s"findMatchInInputAttributes: ${expr}, inputAttributes: ${inputAttributes}")
```

xerial:
Merged.

PR #2598, opened 2022-11-29
airframe-sql: Resolve sub queries in FROM clause

takezoe:


airframe-sql/src/main/scala/wvlet/airframe/sql/analyzer/TypeResolver.scala
```diff
@@ -274,18 +274,20 @@ object TypeResolver extends LogSupport {
       inputAttributes: Seq[Attribute]
   ): List[Expression] = {
     trace(s"findMatchInInputAttributes: ${expr}, inputAttributes: ${inputAttributes}")
+    val resolvedAttributes = inputAttributes.map(resolveAttribute)
```
takezoe, line 277: Resolve `SingleColumn(ResolvedAttribute())` -> `ResolvedAttribute()`

airframe-sql/src/main/scala/wvlet/airframe/sql/model/LogicalPlan.scala
```diff
@@ -161,12 +161,20 @@ trait LogicalPlan extends TreeNode[LogicalPlan] with Product with SQLSig {
...
   def transform(rule: PartialFunction[LogicalPlan, LogicalPlan]): LogicalPlan = {
-    val newNode: LogicalPlan = rule.applyOrElse(this, identity[LogicalPlan])
-    if (newNode.eq(this)) {
-      mapChildren(_.transform(rule))
-    } else {
-      newNode.mapChildren(_.transform(rule))
-    }
+    var curNode: LogicalPlan = this
+    var newNode: LogicalPlan = this
+    // Apply the rule repeatedly until tree gets stabilized
```
takezoe, line 166: Looks like nested subquery is not resolved completely by one iteration, but I'm not sure if applying rules repeatedly until tree is stabilized is the best solution in airframe-sql library.

xerial: `transform` is a generic function for traversing tree nodes and how the tree will be traversed should be consistent. We should apply the transformation only once per node in this `transform` method.


spark-sql has a special traversal method for sub queries: 
https://github.com/apache/spark/blob/a47869af7fa82b708520da123fa0446214f601c2/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/QueryPlan.scala#L456
```
  /**
   * This method is similar to the transform method, but also applies the given partial function
   * also to all the plans in the subqueries of a node. This method is useful when we want
   * to rewrite the whole plan, include its subqueries, in one go.
   */
  def transformWithSubqueries(f: PartialFunction[PlanType, PlanType]): PlanType =
    transformDownWithSubqueries(f)
```

It internally calls `transformUp` method. I think we should add transformUp, which applies the rule recursively to all of the children nodes first, then apply the rule to the node itself.

xerial: transformUp https://github.com/apache/spark/blob/bf5103ae5bb938102850160048e2a1656e648244/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/trees/TreeNode.scala#L557

airframe-sql/src/main/scala/wvlet/airframe/sql/analyzer/TypeResolver.scala
```diff
@@ -274,18 +274,20 @@ object TypeResolver extends LogSupport {
       inputAttributes: Seq[Attribute]
   ): List[Expression] = {
     debug(s"findMatchInInputAttributes: ${expr}, inputAttributes: ${inputAttributes}")
+    val resolvedAttributes = inputAttributes.map(resolveAttribute)
```
takezoe, line 277: Resolve `SingleColumn(ResolvedAttribute())` -> `ResolvedAttribute()`

airframe-sql/src/main/scala/wvlet/airframe/sql/model/LogicalPlan.scala
takezoe: Added `LogicalPlan.transformUp()` and modified `TypeResolver` to use it.

xerial (approved):
👍

PR #2599, opened 2022-11-30
airframe-sql: Set qualifier to columns in CTE

takezoe:


airframe-sql/src/test/scala/wvlet/airframe/sql/analyzer/TypeResolverTest.scala
```diff
@@ -259,30 +259,30 @@ class TypeResolverTest extends AirSpec {
     test("parse WITH statement") {
       val p = analyze("with q1 as (select id from A) select id from q1")
       p.outputAttributes.toList shouldBe List(
-        ResolvedAttribute("id", DataType.LongType, None, Seq(SourceColumn(tableA, a1)), None)
+        ResolvedAttribute("id", DataType.LongType, Some("q1"), Seq(SourceColumn(tableA, a1)), None)
```
takezoe, line 262: Should the qualifier be removed at resolving time in this case?

xerial: Generally speaking, the qualifier should be preserved for handling such cases:
```with q1 as (select id), q2 as (select id)` select q1.id, q2.id from q1, q2```

xerial (approved):
LGTM

airframe-sql/src/test/scala/wvlet/airframe/sql/analyzer/TypeResolverTest.scala
takezoe: Got it!

PR #2607, opened 2022-12-02
airframe-sql: Fix resolveAttribute to keep column qualifier

takezoe:


airframe-sql/src/main/scala/wvlet/airframe/sql/analyzer/TypeResolver.scala
```diff
@@ -254,10 +256,18 @@ object TypeResolver extends LogSupport {
 
   def resolveAttribute(attribute: Attribute): Attribute = {
     attribute match {
-      case SingleColumn(r: ResolvedAttribute, None, _, _) =>
-        r
-      case SingleColumn(r: ResolvedAttribute, Some(alias), _, _) =>
-        r.withAlias(alias)
+      case SingleColumn(r: ResolvedAttribute, alias, qualifier, _) =>
+        r.pipe { attr =>
```
xerial, line 261: Thanks. I didn't know about `pipe` method 😮

PR #2611, opened 2022-12-03
airframe-sql: Refine Union handling

takezoe:
I thought we can handle Union by letting `ResolvedAttribute` has multiple `SourceColumn`s, however it wasn't enough.

This pull request will revert it to represent union columns as a special expression type `MultiSourceColumn` which is similar to `UnionColumn` in the previous PR.


airframe-sql/src/test/scala/wvlet/airframe/sql/analyzer/TypeResolverTest.scala
```diff
@@ -145,39 +145,119 @@ class TypeResolverTest extends AirSpec {
...
+          i.outputAttributes shouldBe List(SingleColumn(MultiColumn(List(ra1, rb1), None), None, None, None))
+        case _ => fail(s"unexpected plan:\n${p.pp}")
+      }
+    }
+
+    test("resolve except") {
+      val p = analyze("select id from A except select id from B") // => Distinct(Except(...))
+      p match {
+        case Distinct(e @ Except(_, _, _), _) =>
+          e.inputAttributes shouldBe List(ra1, ra2) // TODO ra2 shouldn't be included?
```
xerial, line 259: Right. Let's fix it later

xerial:
Thanks. LGTM.

PR #2615, opened 2022-12-05
airframe-sql: Resolve subqueries in expression

takezoe:


xerial:
Thanks. IN sub query is a common expression in segmentation queries.

PR #2616, opened 2022-12-05
airframe-sql: Fix collectExpression() to cover nested plans

takezoe:
`LogicalPlan.resolved` returns `true` even if there are unresolved attributes inside sub queries without this, but this change may have unexpected side-effect?

By the way, `LogicalPlan.resolved` also returns `true` when `Id(col)` remains in the plan. But it looks like impossible to determine which identifiers should be resolved or not systematically on the current mechanism. For example, it should be resolved if it's a column reference, but shouldn't if it's a column or table alias.

Any ideas to correctly detect identifiers that should be resolved?

### DIFF
--- a/airframe-sql/src/main/scala/wvlet/airframe/sql/model/LogicalPlan.scala
+++ b/airframe-sql/src/main/scala/wvlet/airframe/sql/model/LogicalPlan.scala
@@ -40,6 +40,7 @@ trait LogicalPlan extends TreeNode[LogicalPlan] with Product with SQLSig {
     def collectExpression(x: Any): Seq[Expression] = {
       x match {
         case e: Expression  => e :: Nil
+        case p: LogicalPlan => p.expressions
         case Some(x)        => collectExpression(x)
         case s: Iterable[_] => s.flatMap(collectExpression _).toSeq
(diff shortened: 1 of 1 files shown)


xerial:
Oh. LogicalPlan can have sub-expressions. Good catch. This PR itself is good. 

For type resolution, we need to add UnresolvedXXX nodes and convert them to ResolvedXXX with TypeResolver.

- In Spark SQL, all attributes are first marked Unresolved:  https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/unresolved.scala And these unresolved nodes will be resolved using rules defined in https://github.com/apache/spark/blob/3f97cd620a61bc2685bf35215185365e63bedc0a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/Analyzer.scala#L258 
- It seems ZetaSQL uses a similar approach: Mark all nodes resolved https://github.com/google/zetasql/blob/master/docs/resolved_ast.md#resolved-ast-node-hierarchy

xerial:
I think we should add collectChildExpressions to only list up direct child expressions.

PR #2617, opened 2022-12-05
airframe-sql: Resolve HAVING

takezoe:


airframe-sql/src/test/scala/wvlet/airframe/sql/analyzer/TypeResolverTest.scala
```diff
@@ -383,6 +386,30 @@ class TypeResolverTest extends AirSpec {
...
+              _
+            ) =>
+          key1 shouldBe ra1
+          col shouldBe FunctionCall(
+            "count",
+            Seq(AllColumns(None, Some(Seq(tableA)), Some(NodeLocation(1, 12)))),
+            false,
+            None,
+            None,
+            Some(NodeLocation(1, 12))
```
xerial, line 407: @takezoe It's a bit off-topic from this PR. Do you think if AirSpec supports wildcard match inside case class parameters, will it be useful? https://github.com/wvlet/airframe/issues/2619

xerial (approved):
LGTM

PR #2618, opened 2022-12-05
airframe-sql: Implement outputAttributes of Unnest

takezoe:


airframe-sql/src/main/scala/wvlet/airframe/sql/model/LogicalPlan.scala
```diff
@@ -790,9 +790,14 @@ object LogicalPlan {
 
   case class Unnest(columns: Seq[Expression], withOrdinality: Boolean, nodeLocation: Option[NodeLocation])
       extends Relation {
-    override def children: Seq[LogicalPlan]       = Seq.empty
-    override def inputAttributes: Seq[Attribute]  = Seq.empty // TODO
-    override def outputAttributes: Seq[Attribute] = Seq.empty // TODO
+    override def children: Seq[LogicalPlan]      = Seq.empty
+    override def inputAttributes: Seq[Attribute] = Seq.empty // TODO
```
xerial, line 794: ok. Filed #2620

airframe-sql/src/main/scala/wvlet/airframe/sql/model/LogicalPlan.scala
```diff
@@ -790,9 +790,14 @@ object LogicalPlan {
...
+    override def outputAttributes: Seq[Attribute] = {
+      columns.map {
+        case a: ArrayConstructor => SingleColumn(MultiColumn(a.values, None), None, None, a.nodeLocation)
```
xerial, line 797: Not sure whether we should retain all of the input expressions. I think if we can associate the column name for each ArrayConstructor, we can discard Literal values here except the other Attributes (Id, ResolvedAttributes, etc.).

takezoe: That's right. I think the same thing can be applied to SELECT FROM VALUES.

xerial: It's ok to merge. We can fix the behavior later.

takezoe: Got it. Thank you.

PR #2622, opened 2022-12-06
OpenAPI: Generate parameters with default value

### LINKED ISSUE #2596: openapi: default method/object parameters cannot be found in Surface
Only when invoking airframe-http-codegen from sbt-airframe. It looks like a class loader issue. Probably, this is due to the difference of Surface.ofClass(cl) and Surface.of[A]

This is necessary for adding default parameter values https://swagger.io/docs/specification/describing-parameters/

exoego:
Closes #2596 

This PR improved `RuntimeMethodParameter` so it can extract default method parameters defined in trait such as `trait OpenAPIEndpointExample`.
OpenAPIGenerator leverages `RuntimeMethodParameter` and now generates the `default` field in schema.

Note that Scala 3's `StaticMethodParameter` in `CompileTimeSurfaceFactory` is not touched in this PR.
Because this new feature is useful and `httpCodeGen` project is Scala 2 only at this moment, and I am still new to Scala 3's macro 😓

### DIFF
--- a/airframe-http-codegen/src/test/scala/wvlet/airframe/http/openapi/OpenAPITest.scala
+++ b/airframe-http-codegen/src/test/scala/wvlet/airframe/http/openapi/OpenAPITest.scala
@@ -302,7 +302,14 @@ class OpenAPITest extends AirSpec {
         |          in: query
         |          required: true
         |          schema:
-        |            type: string""".stripMargin,
+        |            type: string
+        |            - name: opt1
+        |        - name: p2
+        |          in: query
+        |          required: false
+        |          schema:
+        |            type: string
+        |            default: foo""".stripMargin,
       """  /v1/post1:
         |    post:
         |      summary: post1
--- a/airframe-http-codegen/src/test/scala/wvlet/airframe/http/openapi/OpenAPITest.scala
+++ b/airframe-http-codegen/src/test/scala/wvlet/airframe/http/openapi/OpenAPITest.scala
@@ -543,25 +543,6 @@ class OpenAPITest extends AirSpec {
     // java.awt.Toolkit.getDefaultToolkit.getSystemClipboard
     //      .setContents(new java.awt.datatransfer.StringSelection(yaml), null)
 
-    openapi.paths.get("/v1/get3/{id}").flatMap(_.get("get")).flatMap(_.parameters).flatMap(_.lastOption).map { s =>
-      s match {
-        case s: Parameter              => s.toYAML
-        case OpenAPI.ParameterRef(ref) => s"OpenAPI.ParameterRef($ref)"
-        case _                         => "otherwise"
-      }
-    } shouldBe Some("""name: p2
-        |in: query
-        |required: false
-        |schema:
-        |  type: string
-        |  default: foo
-        |allowEmptyValue: true""".stripMargin)
-
-    ((openapi.components.flatMap(_.schemas).get("OpenAPIEndpointExample.EndpointRequest")) match {
-      case s: OpenAPI.Schema => s.properties.flatMap(_.get("x9"))
-      case _                 => None
-    }) shouldBe Some(Schema("integer", default = None, format = Some("int32")))
-
     fragments.foreach { x =>
       try {
         yaml.contains(x) shouldBe true
--- a/airframe-http-codegen/src/test/scala/wvlet/airframe/http/openapi/OpenAPITest.scala
+++ b/airframe-http-codegen/src/test/scala/wvlet/airframe/http/openapi/OpenAPITest.scala
@@ -13,9 +13,12 @@
  */
 package wvlet.airframe.http.openapi
 import example.openapi.{OpenAPIEndpointExample, OpenAPIRPCExample, OpenAPISmallExample}
+
 import io.swagger.v3.parser.OpenAPIV3Parser
 import wvlet.airframe.http.Router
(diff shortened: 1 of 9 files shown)


exoego:
With

- Java:  v11.0.3
- Scala: v2.13.10

I've looked the generated bytecodse and it looks like that Scala's method parameter with default in trait is encoded into 

```java
public interface OpenAPIEndpointExample {    
    @Endpoint(method = "GET", path = "/v1/get3/:id")
    void get3(final int id, final String p1, final String p2);
    
    public static /* synthetic */ String get3$default$3$(OpenAPIEndpointExample $this) {
        return $this.get3$default$3();
    }

    default public String get3$default$3() {
        return "foo";
    }

    default String get3$default$3() {
        return "foo";
    }
}

public final class OpenAPIEndpointExample$
{
    public static final OpenAPIEndpointExample$ MODULE$ = new OpenAPIEndpointExample$();
    
    private OpenAPIEndpointExample$() {
    }
}
```

However, the current implementation assumes companion objects hold method parameter with default.
https://github.com/wvlet/airframe/blob/49ed35fc42ac1c31e64d82f7d3a9dd8c6d62fc37/airframe-surface/.jvm/src/main/scala/wvlet/airframe/surface/reflect/RuntimeMethodParameter.scala#L70-L74

I think we need to find the method from interface (trait) and invoke it to retrieve default value.
However, since it is an interface, it is unclear `self` in `method.invoke(self)`. We might give `self` that implements the interface.

xerial:
@exoego Thanks for taking a look. Actually, in the compile-time version of Surface, the method owner instance needs to be given to get the default method parameters and generates a lambda function to call the synthetic method: https://github.com/wvlet/airframe/blob/568de2fcf56837833a49ef99a56275519e20ba7d/airframe-surface/src/main/scala-3/wvlet/airframe/surface/CompileTimeSurfaceFactory.scala#L617-L630

For some reason, getMethodArgDefaultValue(instance) might not be used in OpenApiGerator
https://github.com/wvlet/airframe/blob/49ed35fc42ac1c31e64d82f7d3a9dd8c6d62fc37/airframe-surface/.jvm/src/main/scala/wvlet/airframe/surface/reflect/RuntimeMethodParameter.scala#L82-L91


Using annotation is not ideal as we need to duplicate the same value both in the default parameter and the annotation.

exoego:
Scala 2 side is working.

Scala 3 side is the remains.
It is interesting that StaticMethodParameter is used in failing test for Scala 3, unlike Runtime~~ for Scala 2.

airframe-http-codegen/src/test/scala/wvlet/airframe/http/openapi/OpenAPITest.scala
```diff
@@ -145,6 +148,7 @@ class OpenAPITest extends AirSpec {
         |          format: float
         |        x5:
         |          type: number
+        |          default: '1.0'
```
xerial, line 151: 👍

airframe-surface/.jvm/src/test/scala-2/wvlet/airframe/surface/AirSpecBridgeCompat.scala
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package wvlet.airframe.surface
+
+object AirSpecBridgeCompat {
```
xerial, line 16: Note: AirSpec has been released for all Scala {2, 3} x {JVM, JS} combinations, so we no longer need this bridge to munit.

airframe-surface/src/test/scala/wvlet/airframe/surface/MethodSurfaceTest.scala
```diff
@@ -112,4 +116,22 @@ class MethodSurfaceTest extends SurfaceSpec {
...
+    // FIXME: Fix StaticMethodParameter in CompileTimeSurfaceFactory for Scala 3
+    if (!isScalaJS && !isScala3JVM) {
+      assertEquals(h.getDefaultValue, Some("default"))
+
+      val d = new E {
+        override def hello(v: String = "yay"): String = v
+      }
+      val v = h.getMethodArgDefaultValue(d)
+      // Scala.js doesn't support reading default method arguments
+      assertEquals(v, Some("yay"))
```
xerial, line 134: Yay!

xerial:
Scala 3 support still has another blocker at #2449, so no problem if this fix is only for Scala 2 now.

xerial:
> It is interesting that StaticMethodParameter is used in failing test for Scala 3, unlike Runtime~~ for Scala 2.

Runtime reflection (e.g., by using TastyReader) is quite slow, so in Scala 3 I'd like to use compile-time code generation via Scala 3 macros as much as possible. 

If the user (or sbt-airframe) can write a code like `Router.of[Class]`, everything can use only compile-time surface classes.  So asking the user to define a default router for each API might be able to solve this issue. 
```
@RPC
trait MyAPI { ... }

object MyAPI  {
  def router: Router = Router.of[MyAPI] // This generates only compile-time surfaces
}
```

xerial:
Merged the fix. Thanks!

PR #2627, opened 2022-12-07
Disable coloring when using PlainSourceCodeLogFormatter

yuokada:


xerial:
Thanks. By the way, what's the use case? 

Recently, almost all environment supports colored ANSI logs (e.g., less -R), removing ANSI color code with ansi2txt, etc.

yuokada:
Planning to store non-colored logs into a file to send them using fluentd.

xerial:
If that's the case, using a filter plugin like https://github.com/mattheworiordan/fluent-plugin-color-stripper might be useful :) We may still want to see colored logs in an emergency situation.

yuokada:
That plugin doesn't work on `td-agent 4`.

PR #2636, opened 2022-12-16
airframe-sql: Resolve JOIN USING with more than two tables

takezoe:


airframe-sql/src/test/scala/wvlet/airframe/sql/analyzer/TypeResolverTest.scala
```diff
@@ -582,6 +582,30 @@ class TypeResolverTest extends AirSpec {
...
+      joins shouldMatch {
+        case List(
+              JoinOnEq(List(c1, c2), _),
+              JoinOnEq(List(c3, c4), _)
+            ) =>
+          c1 shouldBe ra1.withQualifier("a")
+          c2 shouldBe rb1.withQualifier("b")
+          c3 shouldBe ra2.withQualifier("a")
+          c4 shouldBe rb2.withQualifier("c")
+      }
```
takezoe, line 606:
```sql
SELECT a.id, count(1)
FROM (default.A) AS a 
JOIN (default.B) AS b ON a.id = b.id 
JOIN (default.B) AS c ON a.name = c.name 
GROUP BY a.id HAVING count(1) > 160
```

xerial:
Thanks. Merged

PR #2649, opened 2022-12-17
airframe-sql: Fixes #2646 Resolve AllColumns inputs as ResolvedAttributes

### LINKED ISSUE #2646: airframe-sql: Invalid aggregation key propagation
```
[original]
select name, count(*) cnt from (
  select id, arbitrary(name) name from A
  group by 1
)
group by 1


[resolved]
SELECT arbitrary(name) AS name, count(*) AS cnt   <====== This should be just `name`
FROM (SELECT id, arbitrary(name) AS name FROM default.A GROUP BY id) GROUP BY arbitrary(name) 

[original plan]
[Aggregate]: (id:?, name:?) => (name:?, cnt:?)
  - SingleColumn(Id(name))
  - SingleColumn(FunctionCall(count, AllColumns(*), distinct:false, window:None) as cnt)
  - GroupingKey(Literal(1),Some(NodeLocation(5,10)))
  [Aggregate]:  => (id:?, name:?)
    - SingleColumn(Id(id))
    - SingleColumn(FunctionCall(arbitrary, Id(name), distinct:false, window:None) as name)
    - GroupingKey(Literal(1),Some(NodeLocation(3,12)))
    [TableRef]
      - A

[resolved plan]
[Aggregate]: (id:long, name:?) => (name:?, cnt:?)
  - SingleColumn(SingleColumn(FunctionCall(arbitrary, name:string <- [A.name], distinct:false, window:None) as name))
  - SingleColumn(FunctionCall(count, AllColumns(), distinct:false, window:None) as cnt)
  - GroupingKey(FunctionCall(arbitrary, name:string <- [A.name], distinct:false, window:None),Some(NodeLocation(2,14)))
  [Aggregate]: (id:long, name:string) => (id:long, name:?)
    - SingleColumn(id:long <- [A.id])
    - SingleColumn(FunctionCall(arbitrary, name:string <- [A.name], distinct:false, window:None) as name)
    - GroupingKey(id:long <- [A.id],None)
    [TableScan]:  => (id:long, name:string)
```

xerial:
- Resolve FunctionCall input as a ResolvedAttribute to avoid too eager pull-up of FunctionCall #2646 

An example plan after this fix:
image

- To resolve `arbitrary` aggregation function data type, we need to provide function tables. When no function table is available, resolving its DataType as `name: ? (Unknown type)` is ok. count(...) always returns Long, so added a quick path.
- I think we don't need to propagate detailed expressions like FunctionCall to upper nodes. For example, the following item (iii) can be ResolvedAttribute with source column information:
  1. table scan (source columns: c1, c2, ...)
  2. function call (source columns: c1, c2, ...)
  3. reference to the result of (ii) (source columns: c1, c2, ...)

### DIFF
--- a/airframe-sql/src/main/scala/wvlet/airframe/sql/analyzer/TypeResolver.scala
+++ b/airframe-sql/src/main/scala/wvlet/airframe/sql/analyzer/TypeResolver.scala
@@ -13,7 +13,7 @@
  */
 package wvlet.airframe.sql.analyzer
 import wvlet.airframe.sql.SQLErrorCode
-import wvlet.airframe.sql.analyzer.SQLAnalyzer.{PlanRewriter, Rule}
+import wvlet.airframe.sql.analyzer.RewriteRule.PlanRewriter
 import wvlet.airframe.sql.model.Expression._
 import wvlet.airframe.sql.model.LogicalPlan._
 import wvlet.airframe.sql.model._
@@ -26,19 +26,19 @@ import scala.util.chaining.scalaUtilChainingOps
   */
 object TypeResolver extends LogSupport {
 
-  def typerRules: List[Rule] = {
+  def typerRules: List[RewriteRule] = {
     // First resolve all input table types
     // CTE Table Refs must be resolved before resolving aggregation indexes
-    TypeResolver.resolveCTETableRef _ ::
-      TypeResolver.resolveAggregationIndexes _ ::
-      TypeResolver.resolveAggregationKeys _ ::
-      TypeResolver.resolveSortItemIndexes _ ::
-      TypeResolver.resolveSortItems _ ::
-      TypeResolver.resolveTableRef _ ::
-      TypeResolver.resolveJoinUsing _ ::
-      TypeResolver.resolveSubquery _ ::
-      TypeResolver.resolveRegularRelation _ ::
-      TypeResolver.resolveColumns _ ::
+    TypeResolver.resolveCTETableRef ::
+      TypeResolver.resolveAggregationIndexes ::
+      TypeResolver.resolveAggregationKeys ::
+      TypeResolver.resolveSortItemIndexes ::
+      TypeResolver.resolveSortItems ::
+      TypeResolver.resolveTableRef ::
+      TypeResolver.resolveJoinUsing ::
+      TypeResolver.resolveSubquery ::
+      TypeResolver.resolveRegularRelation ::
+      TypeResolver.resolveColumns ::
       Nil
   }
 
@@ -46,13 +46,14 @@ object TypeResolver extends LogSupport {
     resolve(analyzerContext, plan, typerRules)
   }
 
-  private[sql] def resolve(analyzerContext: AnalyzerContext, plan: LogicalPlan, rules: List[Rule]): LogicalPlan = {
+  private[sql] def resolve(
+      analyzerContext: AnalyzerContext,
+      plan: LogicalPlan,
+      rules: List[RewriteRule]
+  ): LogicalPlan = {
     val resolvedPlan = rules
       .foldLeft(plan) { (targetPlan, rule) =>
-        val r = rule.apply(analyzerContext)
-        // Recursively transform the tree form bottom to up
-        val resolved = targetPlan.transformUp(r)
-        resolved
+        rule.transform(targetPlan, analyzerContext)
       }
     resolvedPlan
   }
@@ -79,22 +80,24 @@ object TypeResolver extends LogSupport {
     * @param context
     * @return
     */
-  def resolveAggregationIndexes(context: AnalyzerContext): PlanRewriter = {
-    case a @ Aggregate(child, selectItems, groupingKeys, having, _) =>
-      val resolvedGroupingKeys: List[GroupingKey] = groupingKeys.map {
-        case k @ GroupingKey(LongLiteral(i, _), _) if i <= selectItems.length =>
-          // Use a simpler form of attributes
-          val keyItem = selectItems(i.toInt - 1) match {
-            case SingleColumn(expr, _, _, _) =>
-              expr
-            case other =>
-              other
-          }
-          GroupingKey(keyItem, k.nodeLocation)
-        case other =>
-          other
-      }
-      Aggregate(child, selectItems, resolvedGroupingKeys, having, a.nodeLocation)
+  object resolveAggregationIndexes extends RewriteRule {
+    def apply(context: AnalyzerContext): PlanRewriter = {
+      case a @ Aggregate(child, selectItems, groupingKeys, having, _) =>
+        val resolvedGroupingKeys: List[GroupingKey] = groupingKeys.map {
+          case k @ GroupingKey(LongLiteral(i, _), _) if i <= selectItems.length =>
+            // Use a simpler form of attributes
+            val keyItem = selectItems(i.toInt - 1) match {
+              case SingleColumn(expr, _, _, _) =>
+                expr
+              case other =>
+                other
+            }
+            GroupingKey(keyItem, k.nodeLocation)
+          case other =>
+            other
+        }
+        Aggregate(child, selectItems, resolvedGroupingKeys, having, a.nodeLocation)
+    }
   }
 
   /**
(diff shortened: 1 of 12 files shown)


airframe-sql/src/main/scala/wvlet/airframe/sql/analyzer/TypeResolver.scala
```diff
@@ -379,7 +385,27 @@ object TypeResolver extends LogSupport {
           case a: Attribute   => a
           case m: MultiColumn => m
           // retain alias for select column
-          case expr => if (isSelectItem) SingleColumn(expr, Some(i.value), None, expr.nodeLocation) else expr
+          case f: FunctionCall =>
+            // Extract source columns
+            val src = Set.newBuilder[SourceColumn]
```
xerial, line 390: SourceColumn has TableCatalog as well, so using Set operation might be a bit heavy.

airframe-sql/src/main/scala/wvlet/airframe/sql/model/Expression.scala
```diff
@@ -570,8 +570,14 @@ object Expression {
       window: Option[Window],
       nodeLocation: Option[NodeLocation]
   ) extends Expression {
-    // TODO: Resolve the function return type using a function catalog
-    // override def dataType: DataType = super.dataType
+    override def dataType: DataType = {
+      if (functionName == "count") {
+        DataType.LongType
```
takezoe, line 575: ❤️

airframe-sql/src/main/scala/wvlet/airframe/sql/analyzer/TypeResolver.scala
```diff
@@ -379,7 +385,27 @@ object TypeResolver extends LogSupport {
           case a: Attribute   => a
           case m: MultiColumn => m
           // retain alias for select column
-          case expr => if (isSelectItem) SingleColumn(expr, Some(i.value), None, expr.nodeLocation) else expr
+          case f: FunctionCall =>
```
takezoe, line 388: Wondering if the same way can be applied to any expression not only `FunctionCall`. 🤔

airframe-sql/src/test/scala/wvlet/airframe/sql/analyzer/TypeResolverTest.scala
```diff
@@ -350,10 +350,17 @@ class TypeResolverTest extends AirSpec {
...
+              _,
+              List(c1, c2 @ SingleColumn(f: FunctionCall, _, _, _)),
+              List(GroupingKey(`ra1`, _)),
+              Some(GreaterThan(col, LongLiteral(10, _), _)),
+              _
+            ) if c1.name == "id" && c2.name == "cnt" && f.functionName == "count" =>
+          f.args shouldMatch { case List(AllColumns(_, Some(cols), _)) =>
+            cols.toSet shouldBe Set(ra1, ra2)
           }
+          col shouldMatch { case ResolvedAttribute("cnt", DataType.LongType, _, sourceColumns, _) => }
```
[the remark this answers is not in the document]
xerial: I think the sourceColumns need to be the all of the column data used for generating this ResolvedAttribute. If we use airframe-sql for tracking data lineage, we may need to maintain this property across SQL expression nodes.

airframe-sql/src/main/scala/wvlet/airframe/sql/analyzer/TypeResolver.scala
xerial: That's also my concern. Maybe converting every type of expression as ResolvedAttribute?

airframe-sql/src/test/scala/wvlet/airframe/sql/analyzer/TypeResolverTest.scala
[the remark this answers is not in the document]
xerial: For example, if the function is max/min/max_by/min_by etc., source column information will be necessary to show the result contains any PID. If the function is count(1), personal information will be lost, so sourceColumns might be unnecessary because raw column values will not be exposed.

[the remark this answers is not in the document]
takezoe: `sourceColumns` seems an empty List in this pull request. Is this expected?

airframe-sql/src/main/scala/wvlet/airframe/sql/analyzer/TypeResolver.scala
takezoe: I'm not 100% but maybe that should be.

airframe-sql/src/test/scala/wvlet/airframe/sql/analyzer/TypeResolverTest.scala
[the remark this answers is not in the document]
xerial: How about defining `sourceColumns` as input columns without any modifications so that data lineage with nested function calls, such as `count(max(..(col)))`, can be traced in a different manner outside TypeResolver? 

If it sounds good, I will fix code and tests accordingly.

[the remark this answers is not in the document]
takezoe: Hmm, I think `TypeResolver` will be unnecessary in that case. Just scanning a tree for any node would work for extracting column lineage.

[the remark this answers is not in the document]
xerial: ok. Let's forget about data lineage. It's not relevant in this PR.

At least, we need to be able to track whether PID/private columns are used as function inputs. 

```scala
          col shouldMatch { case ResolvedAttribute("cnt", DataType.LongType, _, sourceColumns, _) => }
```
this souceColumn should have Set(ra1, ra2)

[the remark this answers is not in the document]
takezoe: We need to give resolved `selectItems` instead of `a.outputAttributes` (= `selectItems`) here.
https://github.com/wvlet/airframe/blob/df40e3f1cc309df963cac8f996f3d2b5b1e885e4/airframe-sql/src/main/scala/wvlet/airframe/sql/analyzer/TypeResolver.scala#L118-L122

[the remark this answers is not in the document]
takezoe: By the way, I noticed that this query is actually invalid on Presto:
```sql
select id, count(*) cnt from A group by id having cnt > 10
```
Got the following error:
```
Query 20221219_122609_00005_r3q5y failed: line 1:84: Column 'cnt' cannot be resolved
```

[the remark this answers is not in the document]
xerial: Good catch 😅

airframe-sql/src/main/scala/wvlet/airframe/sql/analyzer/TypeResolver.scala
xerial: Marked as TODO for now

airframe-sql/src/main/scala/wvlet/airframe/sql/analyzer/TypeResolver.scala
```diff
@@ -108,16 +114,17 @@ object TypeResolver extends LogSupport {
...
           groupingKeys.map(x => {
-            val e = resolveExpression(context, x.child, inputAttributes, false)
+            val e = resolveExpression(context, x.child, childOutputAttributes, false)
             GroupingKey(e, e.nodeLocation)
           })
         val resolvedHaving = having.map {
           _.transformUpExpression { case x: Expression =>
-            resolveExpression(context, x, a.outputAttributes, false)
+            // Having recognize attributes only from the input relation
+            resolveExpression(context, x, childOutputAttributes, false)
```
xerial, line 127: This part had a bug, which was found in 
https://github.com/wvlet/airframe/pull/2649#discussion_r1052424024

airframe-sql/src/test/scala/wvlet/airframe/sql/analyzer/TypeResolverTest.scala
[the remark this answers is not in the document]
xerial: - Fixed https://github.com/wvlet/airframe/pull/2649#discussion_r1052558235
- Make explicit that ResolvedAttribute.sourceColumns is given only when the attribute is a direct reference of table columns without any change (e.g., via function).

airframe-sql/src/main/scala/wvlet/airframe/sql/analyzer/TypeResolver.scala
xerial: Resolved every matched expressions as ResolvedAttribute here

airframe-sql/src/main/scala/wvlet/airframe/sql/model/LogicalPlanPrinter.scala
```diff
@@ -57,7 +57,14 @@ object LogicalPlanPrinter extends LogSupport {
             }
             s": ${printAttr(inputAttrs)} => ${printAttr(outputAttrs)}"
           }
-        val prefix = s"${ws}[${m.modelName}]${functionSig}"
+
+        val prefix = m match {
+          case t: TableScan =>
+            s"${ws}[${m.modelName}] ${t.table.fullName}${functionSig}"
```
xerial, line 63: For showing the input table name after TableRef is resolved to TableScan

airframe-sql/src/main/scala/wvlet/airframe/sql/model/ResolvedPlan.scala
```diff
@@ -97,7 +99,7 @@ case class ResolvedAttribute(
       case (Some(q), columns) if columns.nonEmpty =>
         columns
           .map(_.fullName)
-          .mkString(s"${q},${typeDescription} <- [", ", ", "]")
+          .mkString(s"${q}.${typeDescription} <- [", ", ", "]")
```
xerial, line 102: Fixed a typo

airframe-sql/src/main/scala/wvlet/airframe/sql/analyzer/TypeResolver.scala
```diff
@@ -108,16 +114,17 @@ object TypeResolver extends LogSupport {
...
           groupingKeys.map(x => {
-            val e = resolveExpression(context, x.child, inputAttributes, false)
+            val e = resolveExpression(context, x.child, childOutputAttributes)
             GroupingKey(e, e.nodeLocation)
           })
         val resolvedHaving = having.map {
           _.transformUpExpression { case x: Expression =>
-            resolveExpression(context, x, a.outputAttributes, false)
+            // Having recognize attributes only from the input relation
+            resolveExpression(context, x, childOutputAttributes)
```
xerial, line 127: This needed to use the resolved child attributes

airframe-sql/src/test/scala/wvlet/airframe/sql/analyzer/TypeResolverTest.scala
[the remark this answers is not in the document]
takezoe: > Make explicit that ResolvedAttribute.sourceColumns is given only when the attribute is a direct reference of table columns without any change

Oh, I see.

takezoe (approved):
LGTM

airframe-sql/src/main/scala/wvlet/airframe/sql/analyzer/TypeResolver.scala
```diff
@@ -324,15 +331,14 @@ object TypeResolver extends LogSupport {
   private def findMatchInInputAttributes(
       context: AnalyzerContext,
       expr: Expression,
-      inputAttributes: Seq[Attribute],
-      isSelectItem: Boolean
```
xerial, line 328: MultiColumn is changed to Attribute, so we can use it as SelectItem as is.

Issue #2637, opened 2022-12-16
airspec: Support matching nested test cases


xerial:
```scala
test("A") {
  test("X") { ... }
  test("Y") { ... }
}

# Run A.X but not A.Y
> testOnly -- A.X
```

xerial:
I'll work on it because I need it now :)

PR #2651, opened 2022-12-18
airspec: #2637 Support selecting nested test specs

xerial:
Support selecting nested test cases with `(test name)/(child test name)/...` syntax. 

Could you take a look? @takezoe @exoego

### DIFF
--- a/airspec/src/main/scala/wvlet/airspec/runner/AirSpecContextImpl.scala
+++ b/airspec/src/main/scala/wvlet/airspec/runner/AirSpecContextImpl.scala
@@ -15,6 +15,7 @@ package wvlet.airspec.runner
 
 import wvlet.airframe.Session
 import wvlet.airframe.surface.Surface
+import wvlet.airspec.runner.AirSpecSbtRunner.AirSpecConfig
 import wvlet.airspec.spi.AirSpecContext
 import wvlet.airspec.{AirSpecDef, AirSpecSpi}
 import wvlet.log.LogSupport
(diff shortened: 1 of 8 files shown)


exoego:
I think slash separator is a bit awkward, but may be fine for AirSpec's nesting `test` blocks.

----

AFAIK, MUnit does not support nested tests.
Nested `test` block does not run any assert.


Some styles of ScalaTest supports nested tests.
```scala
class FooSpec extends org.scalatest.funspec.AnyFunSpec {
  describe("A Set") {
    describe("when empty") {
      it("should have size 0") {
        assert(Set.empty.size == 0)
      }

      it("should produce NoSuchElementException when head is invoked") {
        assertThrows[NoSuchElementException] {
          Set.empty.head
        }
      }
    }
  }
}
```
then one can select child test cases with `-z` option and space separators.

```
sbt:cron4s> testOnly -- -z "A Set when empty should have size"
[info] FooSpec:
[info] A Set
[info]   when empty
[info]   - should have size 0
```

I think whitespace separator is more reasonable in `~~Spec` style.

takezoe:
> I think whitespace separator is more reasonable in ~~Spec style since it looks like a natural paragraph.

Interesting. We might want to allow both whitespace and slash as a separator because we often group test cases just by the category name when we write test cases in AirSpec.

xerial:
Thanks.

AirSpec allows partial match of the test names so that the user doesn't always need to specify the exact test names with wildcard(*) like `test name*`. You only need to pick some keywords from individual levels of tests like `(keyword1)/(keyword 2)...`  As test spec names may contain white space characters, I thought having some explicit separator is useful so that we can specify test hierarchies like `"test name/child test name"` without ambiguity. 

ScalaTest -z option is interesting, but if the same keyword appears multiple times like `test A/test A and B` and `test A/test A and C`, `test B/test A and test D`, it's a bit difficult to specify test paths like `testA/*`. -z "test A" will run everything in this example because the keyword "test A" appears in all of the child paths.

xerial:
Let me try this /-slash separator first and figure out if it's better to support other pattern matches like -z :)

Issue #2167, opened 2022-05-10
airframe-http: Use HttpMessage.Request/Response as the standard interface


xerial:
Historically, we needed to manage various http request/response types of Finagle, OkHttp, etc. Now that airframe-http has its own HttpMessage.Request/Response types, we should be able to wrap backend-specific request/response types into the unified http requests and responses.

Converting backend specific requests (e.g., Fingagle request) to HttpMesage.Request is not expensive. Currently, using backend-specific implementation has much higher cost. 

Projects using backend-specific request/response types:
- airframe-http-okhttp
- airframe-http-finagle
- airframe-http-recorder
- airframe-http-grpc

## Benefits:
- We can use the same filters both for finagle and gRPC (e.g., auth filter, header manipulation filter, etc.)
- Simplify the usage of http-recorder, okhttp, etc.

## Steps:
- Add a new factory methods in each projects
- Preserve the current interfaces for a while so as not to break the existing code

xerial:
done

PR #2671, opened 2022-12-26
airframe-http-recorder: Scala 3 support

xerial:
- [x] Use Netty-backend server
- [x] Introduce Rx-based filter #2591 
- [x] Change interface to use HttpMessage.Request/Response #2167


xerial:
@takezoe With this PR, airframe-http-recorder will work in Scala 3 by using the Netty backend server, instead of Finagle. A breaking change is that the interface needs to use HttpMessage.{Request, Response} becaue we can't use Finagle request/response types in Scala 3.

PR #2673, opened 2022-12-27
airframe-sql: Resolve column in nested SELECT *

takezoe:


xerial (approved):
Nice catch. Thanks

xerial:
It seems we have two cases where we want to retain AllColumns(*) and a single column. 

For example, select * from (select * from ...), we want to preserve AllColumns in the input attributes, but for just checking input columns, we need to extract an input column as in this PR. 

Maybe adding an abstraction like AttributeList(inputAttributes: List[Attribute]) might be necessary.

xerial:
For now we have inputColumns and inputAttributes, so we can use them for different purposes.

PR #2678, opened 2022-12-28
airframe-sql: Fix wrong column qualifying in Join

takezoe:


airframe-sql/src/main/scala/wvlet/airframe/sql/model/LogicalPlan.scala
```diff
@@ -691,13 +691,7 @@ object LogicalPlan {
...
     }
     override def inputAttributes: Seq[Attribute] = {
-      (getRelationName(left) match {
-        case Some(q: String) => left.outputAttributes.map(_.withQualifier(q))
-        case None            => left.outputAttributes
-      }) ++ (getRelationName(right) match {
-        case Some(q: String) => right.outputAttributes.map(_.withQualifier(q))
-        case None            => right.outputAttributes
-      })
+      left.outputAttributes ++ right.outputAttributes
```
takezoe, line 694: Qualifying columns here seems to be no longer necessary with the current TypeResolver implementation.

airframe-sql/src/test/scala/wvlet/airframe/sql/parser/SQLGeneratorTest.scala
```diff
@@ -138,4 +151,12 @@ class SQLGeneratorTest extends AirSpec {
     val sql = SQLGenerator.print(resolvedPlan).toLowerCase
     sql shouldBe "select * from (select distinct id from a)"
   }
+
+  test("xxx") {
+    val resolvedPlan =
+      SQLAnalyzer.analyze("select country from (select * from A inner join B using (id))", "default", demoCatalog)
+    println(resolvedPlan.asInstanceOf[Project].selectItems.head)
+    val sql = SQLGenerator.print(resolvedPlan).toLowerCase
+    sql shouldBe "select country from (select * from a join b using (id))"
```
takezoe, line 160: Without this PR, generated SQL is:
```sql
select b.country from (select * from a join b using (id))
```

airframe-sql/src/main/scala/wvlet/airframe/sql/model/LogicalPlan.scala
xerial: 👍

xerial (approved):
Thanks!

PR #2685, opened 2023-01-01
airframe-http-netty: Enable error handling

1hakusai1:
Finagle backend can return custom error responses. (https://wvlet.org/airframe/docs/airframe-http.html#error-handling)
This PR add same feature to Netty backend.


xerial:
Thanks!

PR #2698, opened 2023-01-13
airframe-sql: Support in-expression sub-query in traverse and transform

takezoe:
This will make possible to traverse/transform `LogicalPlan` in `Expression`.

Another idea is moving all traverse and transform methods (including `mapChildren`) to TreeNode with a signature change `LogicalPlan` -> `TreeNode[_]`. This will change the ability of those methods that can traverse/transform any nodes in the tree, not only `LogicalPlan`. Meaning, no need to have methods for `LogicalPlan` and `Expression` independently.

I'm not sure which is better. 🤔

### DIFF
--- a/airframe-sql/src/main/scala/wvlet/airframe/sql/model/Expression.scala
+++ b/airframe-sql/src/main/scala/wvlet/airframe/sql/model/Expression.scala
@@ -52,6 +52,50 @@ sealed trait Expression extends TreeNode[Expression] with Product {
     primaryConstructor.newInstance(args.toArray[AnyRef]: _*).asInstanceOf[Expression]
   }
 
+  def transformPlan(rule: PartialFunction[LogicalPlan, LogicalPlan]): Expression = {
+    def recursiveTransform(arg: Any): AnyRef = {
+      arg match {
+        case e: Expression  => e.transformPlan(rule)
+        case l: LogicalPlan => l.transform(rule)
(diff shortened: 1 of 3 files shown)


airframe-sql/src/main/scala/wvlet/airframe/sql/model/Expression.scala
```diff
@@ -52,6 +52,50 @@ sealed trait Expression extends TreeNode[Expression] with Product {
...
+        case other: AnyRef  => other
+        case null           => null
+      }
+    }
+
+    val newArgs = productIterator.map(recursiveTransform)
+    createInstance(newArgs)
+  }
+
+  def traversePlan[U](rule: PartialFunction[LogicalPlan, U]): Unit = {
```
takezoe, line 71: Simply changing `LogicalPlan.traverse()` to accept `PartialFunction[TreeNode[_], U]` works, but if we do so, `traverse()` will be able to handle `Expression`, not only `LogicalPlan` although `LogicalPlan` already have `traverseExpressions()`.

I think adding `traversePlan()` is more consistent in terms of API design unless we drop methods for `Expression`.

xerial:
LGTM.

I've checked Spark SQL's code, and it seems it also uses different transform methods for Expression and LogicalPlan. It seems merging two types of traversals will be unnecessary.

xerial:
Merged

takezoe:
Thank you!

PR #2702, opened 2023-01-16
airframe-sql: Resolve multiple columns from UNION in sub-query

takezoe:
airframe-sql fails to resolve output columns of UNION if there are multiple columns:
```
[sql]
select id, name from (select id, name from A union all select id, name from B)
[plan]
[Project]: (id:long) => (id:long, name:?)
  - id:long := {*id:long <- A.id, *name:string <- A.name, *id:long <- B.id, *name:string <- B.name}
  - name:? := Id(name)
  [Union]: (id:long, name:string, id:long, name:string) => (id:long)
    [Project]: (id:long, name:string) => (id:long, name:string)
      - *id:long <- A.id
      - *name:string <- A.name
    [Project]: (id:long, name:string) => (id:long, name:string)
      - *id:long <- B.id
      - *name:string <- B.name
      [TableScan] default.B:  => (id:long, name:string)

[unresolved expressions]
name:? := Id(name)
Id(name)
```


xerial (approved):
Nice catch. Thanks

PR #2711, opened 2023-01-21
airframe-sql: Fix outputAttributes of join on using node

xerial:


takezoe (approved):
Thank you!!

xerial:
Instead of using Attribute.inputColumns, added Attribute.outputColumns so as not to expand MultiSourceColumn

PR #2712, opened 2023-01-23
airframe-sql: Fix 'join key not found' error in self-join with USING

takezoe:


airframe-sql/src/main/scala/wvlet/airframe/sql/analyzer/TypeResolver.scala
```diff
@@ -441,9 +441,9 @@ object TypeResolver extends LogSupport {
 
     val results = expr match {
       case i: Identifier =>
-        lookup(i.value).map(toResolvedAttribute(i.value, _)).distinct
+        lookup(i.value).map(toResolvedAttribute(i.value, _))
```
takezoe, line 444: All tests passed but I'm still wondering if this change has any downsides. 🤔

xerial: I think these distinct methods were introduced when we had no MultiSourceColumn.

PR #2717, opened 2023-01-24
airframe-sql: Preserve qualifiers given in SingleColumn

xerial:
- Preserve qualifier in SQL text
- Fix test cases

### DIFF
--- a/airframe-sql/src/main/scala/wvlet/airframe/sql/parser/SQLInterpreter.scala
+++ b/airframe-sql/src/main/scala/wvlet/airframe/sql/parser/SQLInterpreter.scala
@@ -409,7 +409,6 @@ class SQLInterpreter(withNodeLocation: Boolean = true) extends SqlBaseBaseVisito
   }
 
   override def visitExpression(ctx: ExpressionContext): Expression = {
-    trace(s"expr: ${print(ctx)}")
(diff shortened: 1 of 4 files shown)


airframe-sql/src/main/scala/wvlet/airframe/sql/parser/SQLInterpreter.scala
```diff
@@ -399,12 +399,16 @@ class SQLInterpreter(withNodeLocation: Boolean = true) extends SqlBaseBaseVisito
     val alias = Option(ctx.AS())
       .map(x => expression(ctx.identifier()))
       .orElse(Option(ctx.identifier()).map(expression(_)))
-    SingleColumn(expression(ctx.expression()), None, getLocation(ctx))
+    val child = expression(ctx.expression())
+    val qualifier = child match {
+      case a: Attribute => a.qualifier
+      case _            => None
+    }
+    SingleColumn(child, qualifier, getLocation(ctx))
```
xerial, line 407: The true problem was that the qualifier was not properly set when generating LogicalPlan. This fix ensures that presence or absence of qualifiers in select ... columns

takezoe (approved):
Thank you!!

PR #2733, opened 2023-02-03
airframe-launcher: Allow '-' in option name

takezoe:


xerial:
Thanks. I noticed this issue when developing a CLI application ...

PR #2741, opened 2023-02-09
airframe-sql: Fix column aliasing in CTEResolver

takezoe:


airframe-sql/src/test/scala/wvlet/airframe/sql/parser/SQLGeneratorTest.scala
```diff
@@ -171,4 +171,11 @@ class SQLGeneratorTest extends AirSpec {
     val sql = SQLGenerator.print(resolvedPlan).toLowerCase
     sql shouldBe "select * from a as t1"
   }
+
+  test("CTE with column names") {
+    val resolvedPlan =
+      SQLAnalyzer.analyze("with t1 (xid) as (select id as pid from A) select xid from t1", "default", demoCatalog)
+    val sql = SQLGenerator.print(resolvedPlan).toLowerCase
+    sql shouldBe "with t1 as (select id as xid from (select id as pid from a)) select xid from t1"
```
takezoe, line 179: Without this fix, the following SQL (`id as pid as xid` isn invalid) is generated:
```sql
with t1 as (select id as pid as xid from (select id as pid from a)) select xid from t1
```

xerial (approved):
Thanks!

PR #2744, opened 2023-02-09
airframe-sql: Fixes #2743 Unresolve AllColumns before applying TypeResolver

### LINKED ISSUE #2743: airframe-sql: Unresolve AllColumns as a pre-process
After rewriting the input relation, we need to re-resolve AllColumn expression if it already has been resolved. TypeResolver should detect mismatches of resolved columns in AllColumn and the inputAttribute.

xerial:


airframe-sql/src/main/scala/wvlet/airframe/sql/analyzer/TypeResolver.scala
```diff
@@ -24,32 +24,34 @@ import wvlet.log.LogSupport
...
+      resolveColumns ::
+      resolveAggregationIndexes ::
+      resolveAggregationKeys ::
+      resolveSortItemIndexes ::
+      resolveSortItems ::
+      Nil
   }
 
-  private[sql] def resolve(
+  def resolve(
```
xerial, line 48: Exposed this method if some customization (adding rules) is necessary.

takezoe (approved):
Thank you! This should be fine as we have already confirmed that clearing `AllColumns.columns` works.

PR #2763, opened 2023-02-19
airframe-sql: Qualify outputColumns if AllColumns has qualifier

takezoe:
To make it possible to resolve columns outside a sub-query that has qualified `AllColumns` like this:
```sql
select t2.name from A t1 inner join (select * from B) t2 using (id)
```
`t2.name` in the select clause cannot be resolved in this SQL.


xerial (approved):

xerial:
Merged. Thanks

PR #2787, opened 2023-03-13
airspec: Show text diffs upon match failures

xerial:
- Add munit's diff code
- diff test
- airspec: Add text diff


airspec/src/main/scala/wvlet/airspec/internal/diff/MUnitDiff.scala
```diff
@@ -0,0 +1,562 @@
+package wvlet.airspec.internal.diff
+
+import wvlet.airspec.internal.PrettyPrint
+
+import java.util
+import java.util.{Collections, Comparator}
+import scala.collection.JavaConverters._
+
+class Diff(val obtained: String, val expected: String) extends Serializable {
```
exoego, line 9: File name is MUnitDiff.scala. Intended?

xerial: Yes. I just copied the code from MUnit

xerial:
Although this works, the resulting diff format might not be ideal for AirSpec as I'd like to show not only line diffs, but also in-line diffs. And also it adds unmanageable pieces of code taken from mUnit. 

Another option includes https://github.com/eed3si9n/expecty 

Let me close the PR as it will not be merged as is

PR #2795, opened 2023-03-17
airframe-sql: Output columns of qualified AllColumns shouldn't have qualifier

takezoe:


airframe-sql/src/main/scala/wvlet/airframe/sql/analyzer/TypeResolver.scala
```diff
@@ -471,10 +471,10 @@ object TypeResolver extends LogSupport {
           case other =>
             toResolvedAttribute(other.name, other)
         }
-        List(a.copy(columns = Some(qualifier match {
+        List(a.copy(columns = Some((qualifier match {
           case Some(q) => allColumns.filter(_.qualifier.contains(q))
           case None    => allColumns
-        })))
+        }).map(_.withQualifier(None)))))
```
xerial, line 477: I see. Now it makes sense

PR #2803, opened 2023-03-22
airframe-sql: Fix full qualified column name handling

takezoe:


airframe-sql/src/main/scala/wvlet/airframe/sql/model/Expression.scala
```diff
@@ -232,10 +232,10 @@ trait BinaryExpression extends Expression {
 case class ColumnPath(database: Option[String], table: Option[String], columnName: String)
 
 object ColumnPath {
-  def fromQName(contextDatabase: String, fullName: String): Option[ColumnPath] = {
+  def fromQName(fullName: String): Option[ColumnPath] = {
     // TODO Should we handle quotation in the name or just reject such strings?
     fullName.split("\\.").toList match {
-      case List(db, t, c) if db == contextDatabase =>
+      case List(db, t, c) =>
```
takezoe, line 238: To be honest, I'm not sure if this fix is correct.

xerial: Yeah. This fix itself should be ok. 

I think Attribute.matched(ColmnPath) function needs to be implemented properly to check the context (or specified) database properly. https://github.com/wvlet/airframe/blob/a44d5b1d9e4bc7bcf016672e3675df432d20bec1/airframe-sql/src/main/scala/wvlet/airframe/sql/model/Expression.scala#L341. This can be fixed in another PR.

takezoe: It can wrongly match in the current implementation? That's terrible...

PR #2818, opened 2023-04-03
airframe-sql: Support EXTRACT function 

NicolasRichard:
Currently, queries containing `EXTRACT` cannot properly be parsed as `SQLInterpreter` does not override the `visitExtract` method. The default one ends up producing a `null` value when a function call is made inside the `EXTRACT` statement. An exception is later thrown by the parser when it encounters that `null` value.

This PR adds support for the `EXTRACT` functions. It also fixes an issue when generating grouping keys; function call would not be printed correctly.

### DIFF
--- a/airframe-sql/src/main/scala/wvlet/airframe/sql/parser/SQLGenerator.scala
+++ b/airframe-sql/src/main/scala/wvlet/airframe/sql/parser/SQLGenerator.scala
@@ -374,7 +374,7 @@ object SQLGenerator extends LogSupport {
       case l: Literal =>
         l.sqlExpr
       case g: GroupingKey =>
-        g.sqlExpr
+        printExpression(g.child)
       case ParenthesizedExpression(expr, _) =>
         s"(${printExpression(expr)})"
       case a: Alias =>
@@ -412,6 +412,8 @@ object SQLGenerator extends LogSupport {
           }
           .getOrElse("")
         s"${name}(${d}${argList})${wd}"
+      case Extract(interval, expr, _) =>
+        s"EXTRACT(${interval} FROM ${printExpression(expr)})"
       case QName(parts, _) =>
         parts.mkString(".")
       case Cast(expr, tpe, tryCast, _) =>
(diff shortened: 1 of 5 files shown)


airframe-sql/src/test/resources/wvlet/airframe/sql/standard/queries.yml
```diff
@@ -192,3 +192,5 @@
       select 1 as c1
 - sql: |
       select count(1) from t having count(distinct id) > 5
+- sql: | 
+      select EXTRACT(QUARTER FROM FROM_UNIXTIME(time)), COUNT(*) FROM a GROUP BY EXTRACT(QUARTER FROM FROM_UNIXTIME(time))
```
xerial, line 196: Good. Could you add more example queries to have the full coverage? 

Btw, I found GitHub copilot is really good at generating such test queries by looking at the code diff.

NicolasRichard: I added additional tests to cover all intervals. I also pushed a fix that prevented a test from succeeding.

xerial:
Need to reformat the code with `sbt> sql/scalafmtAll`

airframe-sql/src/main/scala/wvlet/airframe/sql/parser/SQLGenerator.scala
```diff
@@ -374,7 +374,7 @@ object SQLGenerator extends LogSupport {
       case l: Literal =>
         l.sqlExpr
       case g: GroupingKey =>
-        g.sqlExpr
+        g.index.getOrElse(printExpression(g.child)).toString
```
xerial, line 377: `g.index.map(_.toString).getOrElse(printExpression(g.child))` is better for the readability.

NicolasRichard: Ahhh 💡. I think I understand how the SQL generation is structured. I reverted the changes to this line and adjusted the implementation of `sqlExpr` for the new `Extract` case class.

xerial:
LGTM now! Thanks for your first contribution! I'll merge it after CI passes

PR #2819, opened 2023-04-04
airframe-sql: Fix equi self-join

NicolasRichard:
Certain queries containing equi self-join cannot be parsed:
```
select A.name
from A
inner join (select name from A) t2 
on A.name = t2.name
```
The type resolver will end up throwing an exception, complaining about ambiguous column references. This is because the source code tries to match columns based on their source table. I guess it's to handle joins with unqualified attributes. 

It took me a while to understand what the code was doing. I think that the matcher should not look at the source table name if a qualifier is present.

### DIFF
--- a/airframe-sql/src/main/scala/wvlet/airframe/sql/model/Expression.scala
+++ b/airframe-sql/src/main/scala/wvlet/airframe/sql/model/Expression.scala
@@ -344,7 +344,8 @@ trait Attribute extends LeafExpression with LogSupport {
       tableName match {
         case Some(tableName) =>
           this match {
-            case r: ResolvedAttribute if r.sourceColumn.exists(_.table.name == tableName) =>
+            case r: ResolvedAttribute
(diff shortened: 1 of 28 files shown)


airframe-sql/src/main/scala/wvlet/airframe/sql/model/Expression.scala
```diff
@@ -344,7 +344,8 @@ trait Attribute extends LeafExpression with LogSupport {
       tableName match {
         case Some(tableName) =>
           this match {
-            case r: ResolvedAttribute if r.sourceColumn.exists(_.table.name == tableName) =>
+            case r: ResolvedAttribute
+                if r.qualifier.orElse(r.sourceColumn.map(_.table).map(_.name)).contains(tableName) =>
```
takezoe, line 348: `.map(_.table).map(_.name)` can be `map(_.table.name)`?

takezoe (approved):
Qualifiers were not considered. Makes sense!

I think this is not self-join specific. So, the title of this PR may not be accurate.

PR #2831, opened 2023-04-11
airframe-sql: Fix concatenation support

NicolasRichard:
The concatenation operator is not currently handled by `SQLInterpreter`. Since `x || a` is equivalent to `CONCAT(x, a)`, we simply map the operator to a function call.


airframe-sql/src/test/resources/wvlet/airframe/sql/standard/queries.yml
```diff
@@ -224,3 +224,5 @@
       select EXTRACT(TIMEZONE_HOUR FROM FROM_UNIXTIME(time)), COUNT(*) FROM a
 - sql: |
       select EXTRACT(timezone_minute FROM FROM_UNIXTIME(time)), COUNT(*) FROM a
+- sql: |
+      select user_agent || 'x', count(*) from impression group by 1
```
xerial, line 228: In case we need to preserve this original syntax, it's better to define a new Expression class like `case class Concatenation(values: Seq[Values]) extends FunctionCall(...)` and add a rule for this in SQLGenerator.

xerial:
LGTM. We can address how to maintain the syntax preservation later.

PR #2836, opened 2023-04-12
fix(airframe-sql): Case should be ignored when resolving identifiers

NicolasRichard:
When resolving types, the library should ignore case as most engines (ex: Trino) do not handle case-sensitive identifiers themselves.


xerial:
Thanks! Merged

Issue #2440, opened 2022-09-22
airframe-http: Deprecate the legacy HttpClient[F[_], Req, Resp] interface


xerial:
This is a proposal to deprecate the current http client interface HttpClient[F[_], Req, Resp] that uses get[Resp], getResource[Resp], post[Resp], etc. New [SyncClient/AsyncClient interfaces](https://github.com/wvlet/airframe/blob/fc9e92c64c94e74fd4e0412a6dd7675fec711279/airframe-http/src/main/scala/wvlet/airframe/http/client/HttpClients.scala#L48) that consolidate these legacy methods into a simple one will be added. For example, you can send a http request `send(request: HttpMessage.Request): HttpMessage.Response` method.

Basic usage:
```scala
import wvlet.airframe.http.Http

val client: SyncClient = Http
  .client
  .withRequestFilter(...)
  .withConnectTimeout(...)
  .newSyncClient(baseUrl)

// Simple request and receives a raw http response
val r: Response = client.send(Http.GET("/path").withHeader(....))

// Send POST request with JSON body and read the response as an MyObj object
client.readAs[MyObj](Http.POST("/path2").withJson(...))

// Send an object data as the request body, and receive the response as ResponseType object
// JSON/MessagePack data will be transformed internally
val resp: ResponseType = client.call[RequestType, ResponseType](Http.POST("/path3"), requestDataObj) 

// When you need to use OkHttp client
import wvlet.airframe.http.okhttp.OkHttp

val okHttpClient: SyncClient = OkHttp
   .client
   .withXXX(...) // OkHttp specific configuration
   .newSyncClient(baseUrl)
```

## What will be changed

- Http clients will use a new SyncClient/AsyncClient interface https://github.com/wvlet/airframe/blob/master/airframe-http/src/main/scala/wvlet/airframe/http/client/HttpClients.scala. Mapping to Scala case classes from HttpResponse are already supported. 
  - Differences of client backend (e.g., Java11 http client, okhttp, Finagle, Netty) will be consumed in HttpChannel interface https://github.com/wvlet/airframe/blob/master/airframe-http/src/main/scala/wvlet/airframe/http/client/HttpChannel.scala
- airframe-finagle, okhttp, recorder interfaces will take the standard HttpMessage.Request/Response. The existing code that use these client implementations need to be adjusted to the new interface, or migrate to the standard Java client, which can be created with Http.client.newSyncClient(...) method. 
- airframe-okhttp, recorder will support Scala 3. 
[...]

PR #2865, opened 2023-04-25
airframe-okhttp: Scala 3 support

xerial:
- Use the standard HttpMessage.Request/Response types for the library interface #2440
- Use the same SyncClient code to use the common retry/circuit breaker logic


xerial:
@shimamoto @takezoe This is a breaking change to support Scala 3 and reuse the same HTTP client code base as much as possible between various clients (Java9 HTTP client, URLConnectionClient, OkHttp client, etc.) The basic usage will be as described in https://github.com/wvlet/airframe/issues/2440

PR #2964, opened 2023-05-17
fix(sql): Improve quotation handling

NicolasRichard:
Quotation marks are not really well handled at the moment. Sometimes, they're stripped, sometimes they're not. Rather than introducing massive changes to preserve the fact that an identifier was originally quotes, I propose the strip them whenever they end up being stored in a `String`. On the generator side, we can than automatically add the quotes back if we detect that they are needed.


airframe-sql/src/main/scala/wvlet/airframe/sql/parser/SQLGenerator.scala
```diff
@@ -379,15 +379,15 @@ object SQLGenerator extends LogSupport {
         s"(${printExpression(expr)})"
       case a: Alias =>
         val e = printExpression(a.expr)
-        s"${e} AS ${a.name}"
+        s"${e} AS ${printNameWithQuotationsIfNeeded(a.name)}"
```
NicolasRichard, line 382: Unfortunately, using `a.sqlExpr` does not work as it's broken (it generates invalid SQL). It would've been nice since, currently, the generation logic is awkwardly happening here and within each expression. 

However, the changes would be way more massive.

airframe-sql/src/main/scala/wvlet/airframe/sql/model/Expression.scala
```diff
@@ -541,6 +545,8 @@ object Expression {
       with LogSupport {
     override def name: String = "*"
 
+    override def sqlExpr: String = name
```
xerial, line 548: qualifier (e.g., `A.*`) can be missing from the generated sql. Is it ok?

NicolasRichard: I removed that change as it ended up not being used. I kept the existing SQL generation logic.

xerial:
LGTM. Merged. Thanks

PR #2983, opened 2023-05-25
fix: Various fixes to make http-recorder usable again. 

NicolasRichard:
As it stands, http-recorder only works with GET requests without a body. This situation is caused by two issues:
- The hash code of the body is not computed correctly. `Arrays.hashcode` should be used to compute an hash code over the content of the array as the `hashcode` method on the array itself only relies on the address of the object.
- `RedirectToRxEndpoint` only exposes a `GET` endpoint. No other verbs can pass-through. I don't know if there's a better way to fix it, but the quick fix I'm proposing works.


airframe-http/src/main/scala/wvlet/airframe/http/router/RedirectToRxEndpoint.scala
```diff
@@ -22,8 +22,31 @@ import wvlet.log.LogSupport
   * @param endpoint
   */
 class RedirectToRxEndpoint(endpoint: RxHttpEndpoint) extends LogSupport {
```
NicolasRichard, line 24: @xerial Any potential side-effects? Any probability something would be relying on the method name or the fact that it only has a GET endpoint?

airframe-http/src/main/scala/wvlet/airframe/http/HttpMessage.scala
```diff
@@ -141,7 +142,7 @@ object HttpMessage {
     def nonEmpty: Boolean = !isEmpty
     def toContentString: String
     def toContentBytes: Array[Byte]
-    def contentHash: Int = toContentBytes.hashCode()
```
xerial, line 144: 😮 Nice catch

PR #2998, opened 2023-06-05
fix: Create new expressions only when needed during transformations

NicolasRichard:
In practice, the arity of an expression is rarely zero. Because of this, `transformUpExpression`/`transformUpExpression` end up creating new objects all the time; even for _identity_ transformation. On its own, it's not really a problem. However, when used in conjunction with `LogicalPlan.transformOnce`, it stops the transform right away and prevent the full traversal of the tree.

In addition, it generates a ton of objects that need to be garbage collected. 
For this reason, the logic should be changed to be closer to what `LogicalPlan` does, i.e. create a new instance if and only if the arguments have changed.


xerial:
Thanks for the optimization. 

This one is a good catch! 
> LogicalPlan.transformOnce, it stops the transform right away and prevent the full traversal of the tree.

PR #3001, opened 2023-06-07
Support sslmode parameter in PostgreSQL

yuokada:
## Overview

- Support `sslmode` parameter in PostgreSQL

## Background

- For relax verification when connecting to a DB using a CNAME record.
- When using a RDS host, a RDS is set up with a SSL certificate managed by AWS. But, a hostname is generated with randomly. It's not convenient. So, some of the users use DNS to set up an arbitrary name and use it to connect to the DB.
- At that case, the default sslmode gets cause of the error since a proper SSL certification isn't used.
- For the `psql` command,  it uses `prefer` when connecting to a DB with ssl. And, it works as expected.

### DIFF
--- a/airframe-jdbc/src/main/scala/wvlet/airframe/jdbc/DbConfig.scala
+++ b/airframe-jdbc/src/main/scala/wvlet/airframe/jdbc/DbConfig.scala
@@ -125,7 +125,7 @@ case class PostgreSQLConfig(
     // SSL configuration for using RDS
     useSSL: Boolean = true,
     sslFactory: String = "org.postgresql.ssl.NonValidatingFactory",
-    sslmode: String = "verify-full"
+    sslmode: String = "prefer"
 )
 
 case class ConnectionPoolConfig(
--- a/airframe-jdbc/src/main/scala/wvlet/airframe/jdbc/DbConfig.scala
+++ b/airframe-jdbc/src/main/scala/wvlet/airframe/jdbc/DbConfig.scala
@@ -124,7 +124,8 @@ case class DbConfig(
 case class PostgreSQLConfig(
     // SSL configuration for using RDS
     useSSL: Boolean = true,
-    sslFactory: String = "org.postgresql.ssl.NonValidatingFactory"
+    sslFactory: String = "org.postgresql.ssl.NonValidatingFactory",
+    sslmode: String = "verify-full"
 )
 
 case class ConnectionPoolConfig(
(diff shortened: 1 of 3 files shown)


airframe-jdbc/src/main/scala/wvlet/airframe/jdbc/DbConfig.scala
```diff
@@ -124,7 +124,8 @@ case class DbConfig(
 case class PostgreSQLConfig(
     // SSL configuration for using RDS
     useSSL: Boolean = true,
-    sslFactory: String = "org.postgresql.ssl.NonValidatingFactory"
+    sslFactory: String = "org.postgresql.ssl.NonValidatingFactory",
+    sslmode: String = "verify-full"
```
yuokada, line 128: "verify-full" is set by default when `ssl` is true.

https://github.com/pgjdbc/pgjdbc/blob/4911ed072681e209423fb608b4bf2daad01bb94d/pgjdbc/src/main/java/org/postgresql/jdbc/SslMode.java#L66-L71

xerial:
Could you explain the background of this change?

yuokada:
@xerial Filled in the background.

airframe-jdbc/src/main/scala/wvlet/airframe/jdbc/DbConfig.scala
xerial: It's a bit confusing because the current default sslFactory is NonValidatingFactory https://github.com/pgjdbc/pgjdbc/blob/master/pgjdbc/src/main/java/org/postgresql/ssl/NonValidatingFactory.java, but the default ssl-mode is verify-full. 

What should be the default values for them?

yuokada: If I understand my issue correctly, the following issue is same. ( The issue was closed without any action. )
https://github.com/pgjdbc/pgjdbc/issues/1459

```
Exception in thread "main" com.zaxxer.hikari.pool.HikariPool$PoolInitializationException: Failed to initialize pool: The hostname ABCDEF.rds.example.com could not be verified by hostnameverifier PgjdbcHostnameVerifier.
```

xerial:
@yuokada Got it. Does this version and verify-full setting work for fixing your issue? If so, let's merge this change and a new version of airframe. 

But it might be better to set the same default values with https://jdbc.postgresql.org/documentation/use/ 
- sslFactory = org.postgresql.ssl.LibPQFactory
- sslMode = prefer 

What do you think?

yuokada:
I confirmed that combination   between ` sslFactory = "org.postgresql.ssl.NonValidatingFactory"` and    `sslmode = "prefer"` works as expected.

Let me only change the default sslMode to `prefer` for backward compatibility.

PR #3007, opened 2023-06-12
airframe-sql: Throw SQLError.RequirementFailed in internal assertions

takezoe:


airframe-sql/src/main/scala/wvlet/airframe/sql/SQLError.scala
```diff
@@ -81,3 +81,18 @@ object SQLErrorCode {
...
   case object InternalError         extends SQLErrorCode(0x10000)
 }
+
+/**
+  * Methods for assertion replacing Predef.require() that throws SQLError instead of IllegalArgumentException.
+  */
+object Assertion {
+  def require(requirement: Boolean, message: => Any, nodeLocation: Option[NodeLocation]): Unit = {
+    if (!requirement)
+      throw SQLErrorCode.InternalError.newException(s"requirement failed: ${message}", nodeLocation)
```
xerial, line 91: InternalError is reserved for retryable errors (e.g., out of memory, system crash, missing subsystems, etc.). This assertion fails every time, so it's better to add a new user error code.

takezoe: Can we use this? (Looks like this error code is not currently used)
https://github.com/wvlet/airframe/blob/70012e85fa2ae8934acc1da4cd2f2e952f668cac/airframe-sql/src/main/scala/wvlet/airframe/sql/SQLError.scala#L70

xerial: Let's add a new user error, which will be mapped to 400 BadRequest in HTTP (UNEXPECTED_STATE_U9, in airframe-rpc):
```scala
  case object RerquirementFailed     extends SQLErrorCode(0x0012) 
```

UserError (0x0000) is a boundary code between Internalerror (0x10000)

takezoe: > UserError (0x0000) is a boundary code between InternalError (0x10000)

I see.

PR #3044, opened 2023-07-04
rx-html (feature): Add RxElement.onMount event hook

### LINKED ISSUE #2988: airframe-rx-http: Add onMount event handler to RxElement
Need a hook that will be called after rendering DOM

xerial:
Closes #2988


xerial:
To see the dom element, registered hooks should run after the top element is mounted to dom.document.

PR #3057, opened 2023-07-11
airframe-sql: Resolve columns from CTE inside AliasedRelation

takezoe:


airframe-sql/src/main/scala/wvlet/airframe/sql/analyzer/TypeResolver.scala
```diff
@@ -380,8 +380,7 @@ object TypeResolver extends LogSupport {
           a.copy(expr = resolved)
         }
       case SingleColumn(a: Attribute, qualifier, _) if a.resolved =>
-        // Optimizes the nested attributes, but preserves qualifier in the parent
-        a.setQualifierIfEmpty(qualifier)
+        a.withQualifier(qualifier)
```
takezoe, line 383: Need to keep the qualifier if `SingleColumn` has it.

airframe-sql/src/test/scala/wvlet/airframe/sql/analyzer/TypeResolverTest.scala
```diff
@@ -1070,4 +1070,19 @@ class TypeResolverTest extends AirSpec with ResolverTestHelper {
...
+      col.fullName shouldBe "id"
+      col.sourceColumn.head.fullName shouldBe "A.id"
+    }
+
+    val p2 = analyze("with t1 as (select id from A) select count(id) from (select id from t1) t2")
+    p2.outputAttributes shouldMatch {
+      case List(SingleColumn(FunctionCall("count", Seq(col: ResolvedAttribute), _, _, _, _), _, _)) =>
+        col.fullName shouldBe "t2.id"
+        col.sourceColumn.head.fullName shouldBe "A.id"
+    }
```
takezoe, line 1086: I'm not sure if adding a qualifier to columns only when column is referred inside function call is expected.

airframe-sql/src/test/scala/wvlet/airframe/sql/analyzer/TypeResolverTest.scala
```diff
@@ -1070,4 +1070,19 @@ class TypeResolverTest extends AirSpec with ResolverTestHelper {
     analyze("select \"prénom\" from (select name as \"prénom\" from A)")
     // No error
   }
+
+  test("resolve CTE in AliasedRelation") {
+    val p1 = analyze("with t1 as (select id from A) select id from (select id from t1) t2")
+    p1.outputAttributes shouldMatch { case List(col: ResolvedAttribute) =>
+      col.fullName shouldBe "id"
```
takezoe, line 1077: I'm not sure if adding a qualifier to columns only when column is referred inside function call is expected.

airframe-sql/src/test/scala/wvlet/airframe/sql/analyzer/TypeResolverTest.scala
```diff
@@ -1070,4 +1070,19 @@ class TypeResolverTest extends AirSpec with ResolverTestHelper {
...
+      col.sourceColumn.head.fullName shouldBe "A.id"
+    }
+
+    val p2 = analyze("with t1 as (select id from A) select count(id) from (select id from t1) t2")
+    p2.outputAttributes shouldMatch {
+      case List(SingleColumn(FunctionCall("count", Seq(col: ResolvedAttribute), _, _, _, _), _, _)) =>
+        col.fullName shouldBe "t2.id"
```
takezoe, line 1084: Without the fix, this was `t1.id` so it couldn't be resolved.

airframe-sql/src/main/scala/wvlet/airframe/sql/analyzer/TypeResolver.scala
xerial: If so, setQualifierIfEmpty method becomes unnecessary (no other usage). Let's remove it

airframe-sql/src/test/scala/wvlet/airframe/sql/analyzer/TypeResolverTest.scala
xerial: In my current understanding, qualifier needs to be added only when the user adds it in the input SQL. If the column cannot be properly resolved due to lack of the qualifier not given in the SQL, there should be other types of bugs in TypeResolver side.

airframe-sql/src/test/scala/wvlet/airframe/sql/analyzer/TypeResolverTest.scala
xerial: I believe the right expectation here is 'id' can be resolved without any qualifier, but the source column is A.id. I guess outputAttributes or some TypeResolver needs to be fixed.

xerial: I think the correct expectation is: 
```scala
col.fullName shouldBe "id" // `t1.id` is wrongly resolved here
```

xerial: Checking the resolved plan with trace-level log:
```
wvlet.airframe.sql.analyzer.TypeResolverTest=trace
```

CTERelationRef already adds t1.id as a qualifier, so I think this part is wrong: 
```
[original plan]
[Query]: (id:?) => (?:long)
  [WithQuery]:  => (id:?)
    - ResolvedIdentifier(Id(t1))
    [Project]:  => (id:?)
      - id:? := Id(id)
      [TableRef]
        - A
  [Project]: (id:?) => (?:long)
    - ?:long := FunctionCall(count, Id(id), distinct:false, window:None)
    [AliasedRelation]:  => (id:?)
      - ResolvedIdentifier(Id(t2))
      [Project]:  => (id:?)
        - id:? := Id(id)
        [TableRef]
          - t1

[resolved plan]
[Query]: (id:long) => (?:long)
  [WithQuery]: (id:long, name:string) => (id:long)
    - ResolvedIdentifier(Id(t1))
    [Project]: (id:long, name:string) => (id:long)
      - *id:long <- A.id
  [Project]: (id:long) => (?:long)
    - ?:long := FunctionCall(count, *t1.id:long <- A.id, distinct:false, window:None)
    [AliasedRelation]: (id:long) => (id:long)
      - ResolvedIdentifier(Id(t2))
      [Project]: (id:long) => (id:long)
        - *id:long <- A.id
        [CTERelationRef]:  => (id:long)
          - *t1.id:long <- A.id
```

xerial: After fixing not to add qualifier to cte.outputAttributes, it seems resolveRegularRelation adds unnecessary qualifier:


```
2023-07-11 14:17:32.139-0700 trace [RewriteRule]
transformed with resolveRegularRelation:
[before]
[Project]: (id:?) => (?:long)
  - ?:long := FunctionCall(count, Id(id), distinct:false, window:None)
  [AliasedRelation]: (id:long) => (id:?)
    - ResolvedIdentifier(Id(t2))
    [Project]: (id:long) => (id:?)
      - id:? := Id(id)
      [CTERelationRef]:  => (id:long)
        - *id:long <- A.id

[after]
[Project]: (id:long) => (?:long)
  - ?:long := FunctionCall(count, *t2.id:long <- A.id, distinct:false, window:None)
  [AliasedRelation]: (id:long) => (id:long)
    - ResolvedIdentifier(Id(t2))
    [Project]: (id:long) => (id:long)
      - id:long := *id:long <- A.id
      [CTERelationRef]:  => (id:long)
        - *id:long <- A.id
  - (RewriteRule.scala:32)
  ```

Comment (xerial) [reply airframe-sql/src/test/scala/wvlet/airframe/sql/analyzer/TypeResolverTest.scala:1084]: Most likely, the current implementation of AliasedRelation.outputAttribuets is wrong, because it adds qualifiers without taking care of the context  https://github.com/wvlet/airframe/blob/dcb0c47375df3e9a50c26d8678361094bfd32092/airframe-sql/src/main/scala/wvlet/airframe/sql/model/LogicalPlan.scala#L518

**Comment by xerial** [comment]
ok. I understand the root cause. Currently, AliasedRelation always returns outputAttributes with qualifiers coming from the table alias name, but this implementation logic is wrong. 

Instead of using the qualifier for resolving columns, which should be used only when the original SQL has it, TypeResolver should understand the context table name alias, say A, and find the referenced columns like A.id.  Attribute or ResolvedAttribute class may need to have `tableNameAlias:Option[String]` parameter. 

Removing withQualifier code in AliasedRelation will be like this https://github.com/wvlet/airframe/commit/345e7dfdd208aa5bfee4eec62d7618e5c7fad28a, but we also need to fix TypeResolver and Attribute classes to pass the unit tests.

**Comment by takezoe** [comment]
`Attribute`s inside `AliasedRelation` need to have qualifiers to resolve them.

**Review by xerial — APPROVED** [review]

File: airframe-sql/src/main/scala/wvlet/airframe/sql/analyzer/TypeResolver.scala [hunk 5]
```diff
@@ -459,20 +452,18 @@ object TypeResolver extends LogSupport {
...
-            a
-          case other =>
-            toResolvedAttribute(other.name, other)
+          // Attribute can be used as is
+          case a: Attribute => a
+          case other        => toResolvedAttribute(other.name, other)
         }
         List(a.copy(columns = Some((qualifier match {
-          case Some(q) => allColumns.filter(_.qualifier.contains(q))
+          case Some(q) => allColumns.filter(c => c.qualifier.contains(q) || c.tableAlias.contains(q))
```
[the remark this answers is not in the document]
takezoe: Rather, `c.qualifier.contains(q)` seems no longer necessary. Will raise another PR.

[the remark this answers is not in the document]
takezoe: https://github.com/wvlet/airframe/pull/3061

PR #3061, opened 2023-07-13
airframe-sql: Remove unnecessary condition from AllColumns resolution

takezoe:
Follow up to https://github.com/wvlet/airframe/pull/3057

### DIFF
--- a/airframe-sql/src/main/scala/wvlet/airframe/sql/analyzer/TypeResolver.scala
+++ b/airframe-sql/src/main/scala/wvlet/airframe/sql/analyzer/TypeResolver.scala
@@ -463,7 +463,7 @@ object TypeResolver extends LogSupport {
           case other        => toResolvedAttribute(other.name, other)
         }
         List(a.copy(columns = Some((qualifier match {
-          case Some(q) => allColumns.filter(c => c.qualifier.contains(q) || c.tableAlias.contains(q))
+          case Some(q) => allColumns.filter(_.tableAlias.contains(q))
           case None    => allColumns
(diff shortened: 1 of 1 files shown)


airframe-sql/src/main/scala/wvlet/airframe/sql/analyzer/TypeResolver.scala
```diff
@@ -463,7 +463,7 @@ object TypeResolver extends LogSupport {
           case other        => toResolvedAttribute(other.name, other)
         }
         List(a.copy(columns = Some((qualifier match {
-          case Some(q) => allColumns.filter(c => c.qualifier.contains(q) || c.tableAlias.contains(q))
+          case Some(q) => allColumns.filter(_.tableAlias.contains(q))
```
xerial, line 466: Good. This is because AllColumn.outputAttributes have both the qualifier and tableAlias

takezoe: > This is because AllColumn.outputAttributes have both the qualifier and tableAlias

This is not accurate explanation.

Without this fix, we needed to check qualifiers for the following case but now this case is covered by `tableAlias`.

```sql
select t1.* from (select t1.time from A) t1
```

On the other hand, checking qualifiers allows to resolve cases like below, but actually this is not a valid SQL which means the current logic can resolve even invalid cases.

```sql
select A.* from (select A.time from A)
select t1.* from (select t1.time from A t1)
```

PR #3068, opened 2023-07-20
airframe-sql: Support cascaded set operations

takezoe:
Currently, `TypeResolver` fails to resolve cascaded set operations like:
```sql
select id from A union all select id from B union all select id from C
```
due to the following error:
> java.lang.IllegalArgumentException: transpose requires all collections have the same size

### DIFF
--- a/airframe-sql/src/main/scala/wvlet/airframe/sql/model/LogicalPlan.scala
+++ b/airframe-sql/src/main/scala/wvlet/airframe/sql/model/LogicalPlan.scala
@@ -778,7 +778,15 @@ object LogicalPlan {
       def collectInputAttributes(rels: Seq[Relation], result: Seq[Seq[Attribute]] = Nil): Seq[Seq[Attribute]] = {
         rels.flatMap {
           case s: SetOperation => collectInputAttributes(s.children, result)
-          case x               => result :+ x.outputAttributes.flatMap(_.inputColumns)
+          case other =>
+            result :+ other.outputAttributes.flatMap {
+              case a: AllColumns => a.inputColumns
+              case other =>
+                other.inputColumns match {
+                  case Seq(i) => Seq(i)
+                  case inputs => Seq(MultiSourceColumn(inputs, None, None, None))
+                }
+            }
         }
       }
 
--- a/airframe-sql/src/main/scala/wvlet/airframe/sql/model/LogicalPlan.scala
+++ b/airframe-sql/src/main/scala/wvlet/airframe/sql/model/LogicalPlan.scala
@@ -13,6 +13,7 @@
  */
 package wvlet.airframe.sql.model
 import wvlet.airframe.sql.analyzer.{QuerySignatureConfig}
+import wvlet.airframe.sql.Assertion._
 import wvlet.log.LogSupport
 
 import java.util.UUID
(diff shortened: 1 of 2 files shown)


airframe-sql/src/main/scala/wvlet/airframe/sql/model/LogicalPlan.scala
```diff
@@ -774,7 +775,21 @@ object LogicalPlan {
...
+      }
+
+      val outputAttributes: Seq[Seq[Attribute]] = collectInputAttributes(children)
+
+      // Verify all relations have the same number of columns
+      require(
+        outputAttributes.map(_.size).distinct.size == 1,
+        "All relations in set operation must have the same number of columns",
+        nodeLocation
+      )
```
takezoe, line 792: To throw assertion error as `SQLError.RequirementFailed`

airframe-sql/src/main/scala/wvlet/airframe/sql/model/LogicalPlan.scala
```diff
@@ -774,7 +775,21 @@ object LogicalPlan {
     override def outputAttributes: Seq[Attribute] = mergeOutputAttributes
     protected def mergeOutputAttributes: Seq[Attribute] = {
       // Collect all input attributes
-      val outputAttributes: Seq[Seq[Attribute]] = children.map(_.outputAttributes.flatMap(_.inputColumns))
+      def collectInputAttributes(rels: Seq[Relation], result: Seq[Seq[Attribute]] = Nil): Seq[Seq[Attribute]] = {
+        rels.flatMap {
+          case s: SetOperation => collectInputAttributes(s.children, result)
+          case x               => result :+ x.outputAttributes.flatMap(_.inputColumns)
+        }
+      }
```
takezoe, line 783: Collect input attributes recursively

airframe-sql/src/main/scala/wvlet/airframe/sql/model/LogicalPlan.scala
```diff
@@ -774,7 +775,21 @@ object LogicalPlan {
     override def outputAttributes: Seq[Attribute] = mergeOutputAttributes
     protected def mergeOutputAttributes: Seq[Attribute] = {
       // Collect all input attributes
-      val outputAttributes: Seq[Seq[Attribute]] = children.map(_.outputAttributes.flatMap(_.inputColumns))
+      def collectInputAttributes(rels: Seq[Relation], result: Seq[Seq[Attribute]] = Nil): Seq[Seq[Attribute]] = {
```
takezoe, line 778: Collect input attributes recursively

airframe-sql/src/main/scala/wvlet/airframe/sql/model/LogicalPlan.scala
```diff
@@ -774,7 +775,21 @@ object LogicalPlan {
...
+        rels.flatMap {
+          case s: SetOperation => collectInputAttributes(s.children, result)
+          case x               => result :+ x.outputAttributes.flatMap(_.inputColumns)
+        }
+      }
+
+      val outputAttributes: Seq[Seq[Attribute]] = collectInputAttributes(children)
+
+      // Verify all relations have the same number of columns
+      require(
```
takezoe, line 788: To throw assertion error as `SQLError.RequirementFailed`

airframe-sql/src/main/scala/wvlet/airframe/sql/model/LogicalPlan.scala
xerial: I don't think this code is recursively collecting input attributes from children nodes, rather it just skips collecting the outputAttributes from the root Set operation node.

A tricky example would be nested set operations like this:
```sql
select id from (select id, x from (select id, x, y from D) union all select id, x from B) union all select id from C?
```
In such a nested case, doesn't it collect attributes from a wrong query plan and report `id, x, y`?

takezoe: Recursively in my original comment may be wrong but can't we expect that children are already resolved when this method is called?

In this case, however, the output attributes become `((A.id, B.id), C.id)` and fail to merge because of lack of multiple input source columns handling. I fixed it and added test cases.

airframe-sql/src/main/scala/wvlet/airframe/sql/model/LogicalPlan.scala
```diff
@@ -774,7 +775,29 @@ object LogicalPlan {
     override def outputAttributes: Seq[Attribute] = mergeOutputAttributes
     protected def mergeOutputAttributes: Seq[Attribute] = {
       // Collect all input attributes
-      val outputAttributes: Seq[Seq[Attribute]] = children.map(_.outputAttributes.flatMap(_.inputColumns))
+      def collectInputAttributes(rels: Seq[Relation], result: Seq[Seq[Attribute]] = Nil): Seq[Seq[Attribute]] = {
+        rels.flatMap {
+          case s: SetOperation => collectInputAttributes(s.children, result)
```
xerial, line 780: btw, which part is setting the result other than Nil? It looks like this argument is unnecessary.

takezoe: Oh, that's right.

PR #3076, opened 2023-07-28
rx-html (feature): Reduce tag/attr name conflicts with SVG

xerial:


airframe-rx-html/src/main/scala/wvlet/airframe/rx/html/Tags.scala
```diff
@@ -170,8 +170,7 @@ trait Tags {
   lazy val summary: HtmlElement = tag("summary")
   // Defines superscripted text
   lazy val sup: HtmlElement = tag("sup")
-  // Defines a container for SVG graphics
-  lazy val svg: HtmlElement = tag("svg")
```
xerial, line 174: svg needs to be scoped in SVG namespace, so this definition is useless

Issue #2891, opened 2023-04-27
airframe-rx: Rx.single may not send any event


xerial:
This happened when implementing async Http clients

xerial:
So far, I couldn't reproduce this issue in #3094. May need another approach to find this issue

PR #3094, opened 2023-08-04
rx (internal): Test the behavior difference of Rx.single (lazy) and Rx.const (immediate)

xerial:
- rx (internal): Test the behavior difference of Rx.single (lazy) and Rx.const (immediate)
- Add test


xerial:
This is a test for https://github.com/wvlet/airframe/issues/2891

PR #3149, opened 2023-08-20
surface (fix): Reduce bytecode size for Scala 3

xerial:
- surface (fix): Reduce methodOf code size for PrimitiveSurface
- debug
- Extract method ref
- Extract classOf expr
- Apply class cast
- Use class.cast


xerial:
Other ideas
- Use externally defined methods for field setter and object builder

PR #3156, opened 2023-08-24
feature: JDK21 Support with Scala 3

xerial:


xerial:
Unfortunately, JDK21 binaries are not yet available for temurin or zulu distributions

PR #3179, opened 2023-08-31
finagle (fix): Properly set http method and path parameters in convertToFinagleRequest

yuokada:
## Overview

- Currently, `convertToFinagleRequest` doesn't respect `Method`, `request path` and etc

### DIFF
--- a/airframe-http-finagle/src/main/scala/wvlet/airframe/http/finagle/package.scala
+++ b/airframe-http-finagle/src/main/scala/wvlet/airframe/http/finagle/package.scala
@@ -179,7 +179,12 @@ package object finagle {
   }
 
   def convertToFinagleRequest(request: HttpMessage.Request): Request = {
-    val req = http.Request()
+    val req = if (request.method == "GET") {
+      val params = request.query.toSeq.map((e) => (e.key, e.value))
+      http.Request(s"${request.path}", params: _*)
+    } else {
+      http.Request(http.Method(request.method), s"${request.path}")
+    }
     for (h <- request.header.entries) {
       req.headerMap.add(h.key, h.value)
     }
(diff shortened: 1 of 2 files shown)


airframe-http-finagle/src/test/scala/wvlet/airframe/http/finagle/FinagleTest.scala
```diff
@@ -64,4 +63,40 @@ class FinagleTest extends AirSpec {
...
+      response.method.name shouldBe "GET"
+      response.path shouldBe "/foo"
+    }
+    test("GET with query parameters") {
+      val req = HttpMessage.Request(HttpMethod.GET, "/foo?bar=true&baz=1234")
+
+      val response = convertToFinagleRequest(req)
+
+      response.method.name shouldBe "GET"
+      response.path shouldBe "/foo"
```
xerial, line 82: I thought the path should contain the query string `?bar=true&baz=1234`. Does it need to be set as params?

airframe-http-finagle/src/main/scala/wvlet/airframe/http/finagle/package.scala
```diff
@@ -179,7 +179,12 @@ package object finagle {
   }
 
   def convertToFinagleRequest(request: HttpMessage.Request): Request = {
-    val req = http.Request()
+    val req = if (request.method == "GET") {
+      val params = request.query.toSeq.map((e) => (e.key, e.value))
```
xerial, line 183:
```suggestion
      val params = request.query.toSeq.map(e => (e.key, e.value))
```
If no type is specified, no need to wrap the arg with paren

airframe-http-finagle/src/test/scala/wvlet/airframe/http/finagle/FinagleTest.scala
yuokada: In my understanding, path doesn't contain a query string.
And, according to the [wikipedia](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier), behavior of the `uri` parameter in the case class, `Request` ,  is different with the definition of URI in the wikipedia.
I was actually confused with it before.

https://github.com/wvlet/airframe/blob/8ef23a39ea5b12f17ac9415abe956275108fb225/airframe-http/src/main/scala/wvlet/airframe/http/HttpMessage.scala#L207-L214

PR #3192, opened 2023-09-05
breaking: Drop JDK8 Support and build for JDK11 or later

### LINKED ISSUE #3191: Drop JDK8 Support
Current known blockers:
- ~AWS EMR, EMR Serverless still require JDK8~  Since EMR 6.11.0 supports JDK17 https://docs.aws.amazon.com/emr/latest/EMR-Serverless-UserGuide/using-java-runtime.html 
- ~Databricks runtime needs JDK8~ https://docs.databricks.com/en/release-notes/runtime/13.3lts.html
  - Note: Spark itself supports the latest JDK
  - jdk11 can be supported by specifying JNAME environment variable https://stackoverflow.com/questions/76021747/how-to-configure-azure-databricks-cluster-to-use-java-11 
　　  - jdk17 is also available with `JNAME=zulu17-ca-amd64` environmental variable

xerial:
I am not sure how long we need to support JDK8 as all known blockers have been resolved #3191, but at the same time, no urgent need exists to drop jdk8 support for a while. We can wait for merging this fix during 2023.

### DIFF
--- a/airspec/build.sbt
+++ b/airspec/build.sbt
@@ -71,7 +71,7 @@ val buildSettings = Seq[Setting[_]](
   crossScalaVersions := targetScalaVersions,
   crossPaths         := true,
   publishMavenStyle  := true,
-  javacOptions ++= Seq("-source", "1.8", "-target", "1.8"),
+  javacOptions ++= Seq("-source", "11", "-target", "11"),
   scalacOptions ++= Seq(
     "-feature",
     "-deprecation"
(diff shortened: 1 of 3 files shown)


xerial:
Spark is also considering dropping Java 11 support in Spark 4.0.0 (https://issues.apache.org/jira/browse/SPARK-44112) and making JDK17 default. Targeting JDK17 may not be such an aggressive goal.

xerial:
Ok. Let's merge this and keep using JDK11 in 23.x series. In the next year, 24.x will use JDK17 or later by default.

PR #3235, opened 2023-09-29
breaking: Drop JDK11 support and make JDK17 the default version

xerial:
Make JDK17 the default version for 24.x series because JDK11 will have no public updates from October 2024. https://endoflife.date/oracle-jdk

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -105,7 +105,7 @@ val buildSettings = Seq[Setting[_]](
       ProblemFilters.exclude[MissingClassProblem]("wvlet.airframe.http.internal.*")
     )
   },
-  javacOptions ++= Seq("--release", "17"),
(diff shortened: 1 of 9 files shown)


build.sbt
```diff
@@ -105,7 +105,7 @@ val buildSettings = Seq[Setting[_]](
       ProblemFilters.exclude[MissingClassProblem]("wvlet.airframe.http.internal.*")
     )
   },
-  javacOptions ++= Seq("-source", "11", "-target", "11"),
+  javacOptions ++= Seq("--release", "17"),
```
xerial, line 108: Still use JDK11 release target until we need to use JDK17-specific features. 
```suggestion
  javacOptions ++= Seq("-source", "11", "-target", "11"),
```

PR #3312, opened 2023-12-19
airframe-sql: Resolve UNNEST earlier to prevent cyclic references

takezoe:
## Case 1

```sql
SELECT id, t.name FROM A CROSS JOIN UNNEST (name) AS t (name)
```

This SQL is valid but fails to resolve with the following error:

```
wvlet.airframe.sql.SQLError: [SyntaxError] line 1:45 name is ambiguous:
- *name:string <- A.name
- *name:?
```

Because the output attribute of AliasedRelation (`t (name)`) is included in the source attributes when resolve `UNNEST (name)` here:

https://github.com/wvlet/airframe/blob/49d31bad29eccd2459eb4c02593d371c1b55082a/airframe-sql/src/main/scala/wvlet/airframe/sql/analyzer/TypeResolver.scala#L319-L321

- `r` is CrossJoin
- `r.inputAttributes` is `left.outputAttributes + right.outputAttributes`
- `right` is AliasedRelation(Unnest)

I think output attributes of AliasedRelation shouldn't be in the scope when Unnest is resolved but I didn't come up with an easy fix for this issue.

## Case 2

```sql
SELECT id, t.name FROM A CROSS JOIN UNNEST (A.name) AS t (name)
```

This SQL is also valid but fails to resolve with the following error:

```
Found unresolved expressions in:
[sql]
SELECT id, t.name FROM A CROSS JOIN UNNEST (A.name) AS t (name)
[plan]
[Project]: (id:long, name:string, name:string) => (id:long, name:?)
  - *id:long <- A.id
  - UnresolvedAttribute(t.name)
  [CrossJoin]: (id:long, name:string, name:string) => (id:long, name:string, name:string)
    - NaturalJoin
    [AliasedRelation]:  => (name:string)
      - ResolvedIdentifier(Id(t))
      [Unnest]:  => (name:string)
        - *A.name:string <- A.name
```

Maybe this is because the resolved UNNEST column has a qualifier. This case might not be UNNEST specific.

### DIFF
--- a/airframe-sql/src/main/scala/wvlet/airframe/sql/analyzer/TypeResolver.scala
+++ b/airframe-sql/src/main/scala/wvlet/airframe/sql/analyzer/TypeResolver.scala
@@ -38,6 +38,7 @@ object TypeResolver extends LogSupport {
       resolveTableRef ::
       resolveJoinUsing ::
       resolveSubquery ::
+      resolveJoinUnnest ::
       resolveRegularRelation ::
       resolveColumns ::
       resolveAggregationIndexes ::
@@ -231,6 +232,22 @@ object TypeResolver extends LogSupport {
     }
   }
 
+  object resolveJoinUnnest extends RewriteRule {
+    override def apply(context: AnalyzerContext): PlanRewriter = {
+      case j @ Join(_, _, a @ AliasedRelation(u: Unnest, _, _, _), _, _) =>
+        j.copy(right = a.copy(child = resolveUnnest(u, j, context)))
+      case j @ Join(_, a @ AliasedRelation(u: Unnest, _, _, _), _, _, _) =>
+        j.copy(left = a.copy(child = resolveUnnest(u, j, context)))
+    }
+
+    private def resolveUnnest(u: Unnest, j: Join, context: AnalyzerContext): Unnest = {
+      val resolvedAttributes = j.outputAttributes.filter(_.resolved)
+      u.transformUpExpressions { case x: Identifier =>
+        resolveExpression(context, x, resolvedAttributes)
+      }.asInstanceOf[Unnest]
+    }
+  }
+
   object resolveJoinUsing extends RewriteRule {
     def apply(context: AnalyzerContext): PlanRewriter = {
       case j @ Join(joinType, left, right, u @ JoinUsing(joinKeys, _), _) =>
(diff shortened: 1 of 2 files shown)


xerial:
I also don't have a clear idea of how we should represent Unnest. Need to check how UNNEST is handled in Trino or SparkSQL.

takezoe:
- Spark doesn't seem to support UNNEST (It has EXPLODE function instead)
- Trino seems to create an analysis for UNNEST first. This prevents cyclic references in UNNEST expression

airframe-sql/src/main/scala/wvlet/airframe/sql/analyzer/TypeResolver.scala
```diff
@@ -375,8 +392,8 @@ object TypeResolver extends LogSupport {
         } else {
           a.copy(expr = resolved)
         }
-      case SingleColumn(a: Attribute, qualifier, _, _) if a.resolved =>
-        a
+      case SingleColumn(a: Attribute, qualifier, tableAlias, _) if a.resolved =>
+        a.withQualifier(qualifier).withTableAlias(tableAlias)
```
takezoe, line 396: The following case cannot be resolved without this fix:
```sql
select t.id from (select id from A) as t(id)
```

airframe-sql/src/main/scala/wvlet/airframe/sql/analyzer/TypeResolver.scala
```diff
@@ -231,6 +232,22 @@ object TypeResolver extends LogSupport {
...
+  object resolveJoinUnnest extends RewriteRule {
+    override def apply(context: AnalyzerContext): PlanRewriter = {
+      case j @ Join(_, _, a @ AliasedRelation(u: Unnest, _, _, _), _, _) =>
+        j.copy(right = a.copy(child = resolveUnnest(u, j, context)))
+      case j @ Join(_, a @ AliasedRelation(u: Unnest, _, _, _), _, _, _) =>
+        j.copy(left = a.copy(child = resolveUnnest(u, j, context)))
+    }
+
+    private def resolveUnnest(u: Unnest, j: Join, context: AnalyzerContext): Unnest = {
+      val resolvedAttributes = j.outputAttributes.filter(_.resolved)
```
takezoe, line 244: Excluding unresolved attributes is a workaround to prevent cyclic references.

xerial:
Ok. Thanks for the explanation. I understand the fix

PR #3319, opened 2023-12-23
surface (fix): Support Scala 3 opaque type aliases with type args

xerial:


xerial:
It seems adding a new parameter to the Alias surface introduces binary incompatibilities, especially in Scala.js. An alternative approach need to be tested.

Issue #3419, opened 2024-02-27
Unable to get a surface of a class with a protected constructor in Scala 3


OndrejSpanel:
airframe-surface ver. 24.2.2, Scala 3.3.2

```
import wvlet.airframe.surface.Surface

import scala.collection.immutable.Queue

object Main {
  case class S(q: Queue[String])

  Surface.of[S]

  def main(args: Array[String]): Unit = {

  }
}
```

Fails with error 

> constructor Queue cannot be accessed as a member of scala.collection.immutable.Queue[String] from module class Main$.
>  Access to protected constructor Queue not permitted because enclosing object Main
>  is not a subclass of class Queue in package scala.collection.immutable where target is defined

Reduced repro:

```scala
  class Q protected {}

  Surface.of[Q]
```

There is no issue when the constructor is private.

OndrejSpanel:
Package private constructor has the same issue:

```
  object Obj {
    class Q private[Obj] {}
  }

  Surface.of[Obj.Q]
```

> constructor Q cannot be accessed as a member of Main.Obj.Q from module class Main$.

PR #3423, opened 2024-02-27
surface (fix): Fix #3416 and #3419 - protected and package private access

### LINKED ISSUE #3416: Unable to get a surface of a class with package private members in Scala 3
Similar to #3356, tested with `"org.wvlet.airframe" %% "airframe-surface" % "24.2.2"`

Unable to generate surface of a class with package private values

```scala
import wvlet.airframe.surface.Surface

object Tools {
  class Pos(private[Tools] val pos: String) {
    private[Tools] def getPos: String = pos
  }
}

object Main {
  Surface.of[Tools.Pos]

  def main(args: Array[String]): Unit = {

  }
}
```

Error is:

> value pos cannot be accessed as a member of (x : Tools.Pos) from module class Main$.

Notes: 
- `Surface.methodsOf[Tools.Pos]` gives the same error.
- when I remove `getPos`, there are no errors when calling `Surface.methodsOf[tools.Tools.Pos]` 
- when I make `pos` public, I get a different error with `methodsOf`:

```scala
object Tools {
  class Pos(val pos: String) {
    private[Tools] def getPos: String = pos
  }
}

object Main {
  Surface.methodsOf[Tools.Pos]

  def main(args: Array[String]): Unit = {

  }
}
```

Error is:

> method getPos cannot be accessed as a member of Tools.Pos from module class Main$.

All these variants work fine in Scala 2.

OndrejSpanel:
Starting with tests - I hope I will be able to implement the fix as well.

### DIFF
--- a/airframe-surface/src/main/scala-3/wvlet/airframe/surface/CompileTimeSurfaceFactory.scala
+++ b/airframe-surface/src/main/scala-3/wvlet/airframe/surface/CompileTimeSurfaceFactory.scala
@@ -598,11 +598,12 @@ private[surface] class CompileTimeSurfaceFactory[Q <: Quotes](using quotes: Q):
       // Generate a field accessor { (x:Any) => x.asInstanceOf[A].(field name) }
       val paramIsAccessible =
         t.typeSymbol.fieldMember(paramName) match
-          case nt if nt == Symbol.noSymbol      => false
-          case m if m.flags.is(Flags.Private)   => false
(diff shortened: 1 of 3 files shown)


OndrejSpanel:
All tests added have already passed in Scala 2. After the fix they pass in Scala 3 as well.

xerial:
@OndrejSpanel Thanks for the fix! I've applied scalafmtAll to fix the code format. Otherwise, it looks good to me.

Issue #3411, opened 2024-02-25
Surface.methodsOf for Enumeration fails in Scala 3


OndrejSpanel:
Another one discovered thanks to using the workaround suggested in #3367. Tested with Scala 3.4.1-RC1 and surface 24.2.2, works fine in Scala 2:

build.sbt:

```
version := "0.1.0-SNAPSHOT"

scalaVersion := "3.4.1-RC1"

name := "airframe-surface-test"

libraryDependencies += "org.wvlet.airframe" %% "airframe-surface" % "24.2.2"

```

Main.scala:
```
import wvlet.airframe.surface.Surface

object Main {

  object ScalaEnum extends Enumeration {
    val EnumA, EnumB = Value

    type ScalaEnum = Value
  }

  def main(args: Array[String]): Unit = {
    Surface.methodsOf[ScalaEnum.ScalaEnum]
  }
}
```

produces following error:

```text
C:\Dev\airframe-surface-test\src\main\scala\Main.scala:12:22
constructor ValueSet cannot be accessed as a member of Enumeration.this.ValueSet from object Main.
  private[ValueSet] constructor ValueSet can only be accessed from class ValueSet in class Enumeration.
    Surface.methodsOf[ScalaEnum.ScalaEnum]

```

OndrejSpanel:
I guess with #3423 the behaviour will change, as `private[ValueSet]` should be handled gracefully. It might be worth checking what is the current behaviour with Scala 2 and make sure with Scala 3 it behaves the same way.

PR #3429, opened 2024-02-28
Test for Enumeration support (#3411)

OndrejSpanel:
Test for #3411

With Scala 3 the test currently fails with:

```
[error] -- [E007] Type Mismatch Error: C:\Dev\airframe\airframe-surface\src\test\scala\wvlet\airframe\surface\i3411.scala:30:29
[error] 30 |    val m = Surface.methodsOf[SomeEnum]
[error]    |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^
[error]    |    Found:    Enumeration.this.Value
[error]    |    Required: wvlet.airframe.surface.i3411.SomeEnum.Value
```


OndrejSpanel:
The same error can be seen with a standalone repro:

```scala
  trait Base {
    class InnerType {
      def compare(that: InnerType): Int = 0
    }
  }

  object OuterType extends Base {
    def create: InnerType = new InnerType
  }

    val m = Surface.methodsOf[OuterType.InnerType]

```

```
val m = Surface.methodsOf[OuterType.InnerType]
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Found:    Base.this.InnerType
Required: wvlet.airframe.surface.i3411.OuterType.InnerType
```

In this case the type of the compare parameter is represented as:

> TypeRef(ThisType(TypeRef(ThisType(TypeRef(ThisType(TypeRef(NoPrefix,module class surface)),module class i3411$)),trait Base)),class InnerType)

When using `.show` on this, I get `Base.this.InnerType`. I would like to omit the `Base` from the path. I will inspect the corresponding `TypeRef` a hopefully I can find a pattern.

OndrejSpanel:
With a significant effort I was able to manipulate the `TypeRef` so that I can get `OuterType.InnerType` instead of `Base.this.InnerType`. However when I do this, I get `NoDenotation.owner` later on.

I am afraid this might be too difficult for me - there are too many unfamiliar concepts and too little documentation.

OndrejSpanel:
> However when I do this, I get `NoDenotation.owner` later on.

### Possible Scala 3 bug

This was due to a mistake on my part. I was able to create the correct type, yet in the result the type is interpreted as `Base.this.InnerType`. I start to suspect a bug on Scala 3 side.

I think the type should be `OuterType.InnerType` from the beginning, there is no reason to get `Base.this.InnerType` at all. I will try to check the process step by step and report if I find anything.

### Workaround?

As for supporting `Enumeration`, perhaps the more practical solution would be to implement some workaround. At least in my case I am not interested in methods of `Enumeration.Value` at all, but it is important that it does not fail when an `Enumeration` is present somewhere in AST. This would not solve the `OuterType` / `InnerType` underlying cause, but that is easier to avoid by adjusting the AST sources - in case of `Enumeration` there is no choice and the path dependent `Value` is always there.

OndrejSpanel:
Note: it seems the tests are currently failing with Scala 2 / Scala.js:

> [error] /home/runner/work/airframe/airframe/airframe-surface/src/test/scala/wvlet/airframe/surface/i3411.scala:30:30: type mismatch;
> [error]  found   : Enumeration.this.Value
> [error]  required: wvlet.airframe.surface.i3411.SomeEnum.Value
> [error]     val m = Surface.methodsOf[SomeEnum]

The error is very similar to the one provided by Scala 3. The Scala 2 / JVM passes fine. I do not understand what this means.

xerial:
@OndrejSpanel Thank you for leaving the work log. I'm also not worried about supporting the legacy enumeration as Scala 3 enum is already supported. If we can just get Surface.methodsOf to compile, ignoring these enumeration methods that use dependent types is ok.

Issue #3433, opened 2024-03-01
Cannot get `methodsOf` for some generic types, esp. in Scala 3


OndrejSpanel:
Using following source file all tests fail in Scala 3, and test with `ScriptingContext[_]` fails even with Scala 2:

```scala
  type EvalFunction1[T] = scala.Function1[T, Any]

  class ScriptingContext[X] {
    def func(): Int = ???
    def `then`(f: EvalFunction1[X]): ScriptingContext[Any] = ???
  }

  class SC extends ScriptingContext[Int]

  test("Support generic classes") {
    Surface.of[ScriptingContext[Int]]
    val methods = Surface.methodsOf[ScriptingContext[Int]]
    val methodNames = methods.map(_.name)
    methodNames shouldContain "func"
    methodNames shouldContain "then"
  }

  test("Support generic classes 2") {
    Surface.of[SC]
    val methods = Surface.methodsOf[SC]
    val methodNames = methods.map(_.name)
    methodNames shouldContain "func"
    methodNames shouldContain "then"
  }

  test("Support generic classes 3") {
    Surface.of[ScriptingContext[_]]
    val methods = Surface.methodsOf[ScriptingContext[_]]
    val methodNames = methods.map(_.name)
    methodNames shouldContain "func"
    methodNames shouldContain "then"
  }

```

Scala 3 compile error are:

> Found:    wvlet.airframe.surface.i3418.EvalFunction1[X]
> Required: wvlet.airframe.surface.i3418.EvalFunction1[Int]

And:

> Found:    wvlet.airframe.surface.i3418.EvalFunction1[X]
> Required: wvlet.airframe.surface.i3418.EvalFunction1[
>   wvlet.airframe.surface.i3418.ScriptingContext[?]#X]

OndrejSpanel:
This is in principle a lot similar to #3411

I start wondering: where exactly is the type mismatch happening? It would not be that bad if inexact method parameter was returned, but the problem is it is a compilation error, failing the project build completely as soon as any such method exists, even if the user is not interested in that particular method.

PR #3440, opened 2024-03-02
Surface - Method type exploration

OndrejSpanel:
This an exploratory PR for #3433 and #3429

I would like to know where exactly is the type mismatch happening and if perhaps there is a way to avoid it completely, even at cost of reported method surfaces being not 100% accurate.

### DIFF
--- a/airframe-surface/src/main/scala-3/wvlet/airframe/surface/CompileTimeSurfaceFactory.scala
+++ b/airframe-surface/src/main/scala-3/wvlet/airframe/surface/CompileTimeSurfaceFactory.scala
@@ -863,18 +863,23 @@ private[surface] class CompileTimeSurfaceFactory[Q <: Quotes](using quotes: Q):
           name != "<init>"
         }
 
-    allMethods.filter(m => isOwnedByTargetClass(m, t))
+    allMethods.filter(m => isOwnedByTargetClass(m, t) && !enumerationWorkaround(m, t))
 
   private def nonObject(x: Symbol): Boolean =
     !x.flags.is(Flags.Synthetic) &&
       !x.flags.is(Flags.Artifact) &&
       x.fullName != "scala.Any" &&
       x.fullName != "java.lang.Object" &&
-      // Exclude caes class methods
+      // Exclude case class methods
       x.fullName != "scala.Product"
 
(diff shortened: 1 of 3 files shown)


OndrejSpanel:
Logging says:

> wvlet.airframe.surface.ClassMethodSurface.apply(1, __s000, "compare", wvlet.airframe.surface.Primitive.Int, wvlet.airframe.surface.StaticMethodParameter.apply(method = wvlet.airframe.surface.MethodRef.apply(owner = classOf[wvlet.airframe.surface.i3439.OuterType.InnerType], name = "compare", paramTypes = classOf[Base.this.InnerType], isConstructor = false), index = 0, name = "that", isRequired = false, isSecret = false, surface = __s001, defaultValue = scala.None, accessor = scala.None, methodArgAccessor = scala.None).toIndexedSeq, scala.Some.apply[scala.Function2[scala.Any, scala.collection.immutable.Seq[scala.Any], scala.Any]](((x: scala.Any, `argsÎ“Ã©Ã¢`: scala.collection.immutable.Seq[scala.Any]) => x.asInstanceOf[wvlet.airframe.surface.i3439.OuterType.InnerType].compare(`argsÎ“Ã©Ã¢`.apply(0).asInstanceOf[Base.this.InnerType]))))

Notice last part: `x.asInstanceOf[wvlet.airframe.surface.i3439.OuterType.InnerType].compare(`argsÎ“Ã©Ã¢`.apply(0).asInstanceOf[Base.this.InnerType]))))`

Method `compare` is called, but the argument is of a wrong type. This seems to be a `methodCaller` field. It is a pity to get a mismatch here, because the external signature is `(Any, Seq[Any]) => Any`.

When I disable `methodCaller` creation by replacing it with `'{ Option.empty[(Any, Seq[Any]) => Any] }`, I get a compiler exception someplace else:

"exception while retyping Base.this of class This # -1"

OndrejSpanel:
I have decided not to explore this any more. For my purposes I do not need the `objectFactory` or `methodCaller` functionality, therefore I decided to create a fork of the library instead without this functionality. The fork has no other ambition than to serve as  a temporary workaround until the issues are fixed here.

If anyone is interested, the fork can be found at https://github.com/OpenGrabeso/light-surface. I intend to publish the artifacts in GitHub Packages.

PR #3451, opened 2024-03-13
Surface: __sXXX is a forward reference extending over the definition of __sYYY in methodsOf

OndrejSpanel:
In some complex classes I get errors like "__s239 is a forward reference extending over the definition of __s041" when using `methodsOf` in Scala 3 on them.

This is because `CompileTimeSurfaceFactory.methodsOf` lists the methods in the `seen` order. Some of them are declared as `val`, some as `lazy val`. While `lazy val` can reference each other freely, they cannot do so across a barrier of `val`.

This change makes sure the `lazy val`s (non-lazy surfaces) are listed first.

I have verified this fixes the issue for my complex class (about 400 members). I will try to create a standalone open-source repro later if possible - this will probably need some mixture of methods with lazy and non-lazy surfaces used for types.

I did not check Scala 2 implementation as I am not using Surface on Scala 2 - it is possible it experiences the same issue. If there is a test, this should show up.

### DIFF
--- a/airframe-surface/src/main/scala-3/wvlet/airframe/surface/CompileTimeSurfaceFactory.scala
+++ b/airframe-surface/src/main/scala-3/wvlet/airframe/surface/CompileTimeSurfaceFactory.scala
@@ -748,7 +748,7 @@ private[surface] class CompileTimeSurfaceFactory[Q <: Quotes](using quotes: Q):
       methodsOfInternal(t).asTerm
     ).asExprOf[Seq[MethodSurface]]
 
-    println(s"===  methodOf: ${t.typeSymbol.fullName} => \n${expr.show}")
+    // println(s"===  methodOf: ${t.typeSymbol.fullName} => \n${expr.show}")
     expr
 
   private val seenMethodParent = scala.collection.mutable.Set[TypeRepr]()


OndrejSpanel:
> I will try to create a standalone open-source repro

> Surface on Scala 2 - it is possible it experiences the same issue. If there is a test, this should show up.

The repro is done. It fails before the fix - and it fails on Scala 2.13. I will try to implement a Scala 2 fix as well.

OndrejSpanel:
Scala 2 failure was different, it was because its methods contain `productElementNames`, which I did not expect. Other than that, it seems to work fine (at least with this repro). I have relaxed the test and it is passing now.

OndrejSpanel:
I think the fix has introduced issues for some other complex classes. I will need more time to check.

airframe-surface/src/main/scala-3/wvlet/airframe/surface/CompileTimeSurfaceFactory.scala
```diff
@@ -745,7 +748,7 @@ private[surface] class CompileTimeSurfaceFactory[Q <: Quotes](using quotes: Q):
       methodsOfInternal(t).asTerm
     ).asExprOf[Seq[MethodSurface]]
 
-    // println(s"===  methodOf: ${t.typeSymbol.fullName} => \n${expr.show}")
+    println(s"===  methodOf: ${t.typeSymbol.fullName} => \n${expr.show}")
```
xerial, line 751:
```suggestion
    // println(s"===  methodOf: ${t.typeSymbol.fullName} => \n${expr.show}")
```

PR #3455, opened 2024-03-18
surface: Fix #3454 - methods in generic classes - for Scala 3

### LINKED ISSUE #3454: Surface: Cannot get `methodsOf` for simple generic types
Following code does not compile with Scala 3:

```scala
  case class Gen[X](value: X) {
    def pass(x: X): X = x
  }

  test("Methods of generic type") {
    val methods = Surface.methodsOf[Gen[String]]
  }
```

The error is:

```
[error] -- [E007] Type Mismatch Error: C:\Dev\airframe\airframe-surface\src\test\scala\wvlet\airframe\surface\GenericMethodTest.scala:44:35
[error] 44 |    val methods = Surface.methodsOf[Gen[String]]
[error]    |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[error]    |    Found:    X
[error]    |    Required: String
```

Note: In my ["forked" library ](https://github.com/OpenGrabeso/light-surface) version the same call does not fail, yet the method surfaces are not correct. Both  `returnType` and type of parameter 'x' of `pass` is not `String` but `Object`. Maybe type arguments are not applied for the methods when acquiring their surface.

With Scala 2 the code compiles, but the returned types are not applied, similar to my forked library. Following test can be compiled on Scala 2 but it fails:

```scala
  case class Gen[X](value: X) {
    def pass(x: X): X = x
    def myself: Gen[X] = this
    def wrap(x: X): Gen[X] = Gen[X](value)
    def unwrap(x: Gen[X]): X = x.value
  }

  test("Methods of generic type") {
    val typeSurface = Surface.of[Gen[String]]
    val methods = Surface.methodsOf[Gen[String]]
    val pass = methods.find(_.name == "pass").get
    pass.returnType shouldBe Surface.of[String]
    val myself = methods.find(_.name == "myself").get
    myself.returnType shouldBe typeSurface
    val wrap = methods.find(_.name == "wrap").get
    wrap.returnType shouldBe typeSurface
    val unwrap = methods.find(_.name == "unwrap").get
    unwrap.returnType shouldBe Surface.of[String]
  }

```

OndrejSpanel:
This PR implements a test and Scala 3 fix of #3454. The major part of the fix is a simplification of `typeMappingTable` function. I am not 100% sure what the previous implementation using `method.paramSymss` was trying to do, I only hope whatever it was, it had tests in place for that.

@xerial The new test fails on Scala 2 - if you or anyone is interested in completing the fix, feel free to continue.

### DIFF
--- a/airframe-surface/src/main/scala-3/wvlet/airframe/surface/CompileTimeSurfaceFactory.scala
+++ b/airframe-surface/src/main/scala-3/wvlet/airframe/surface/CompileTimeSurfaceFactory.scala
@@ -305,16 +305,17 @@ private[surface] class CompileTimeSurfaceFactory[Q <: Quotes](using quotes: Q):
       }
   }
 
+  // Build a table for resolving type parameters, e.g., class MyClass[A, B]  -> Map("A" -> TypeRepr, "B" -> TypeRepr)
   private def typeMappingTable(t: TypeRepr, method: Symbol): Map[String, TypeRepr] =
     val classTypeParams = t.typeSymbol.typeMembers.filter(_.isTypeParam)
     val classTypeArgs: List[TypeRepr] = t match
       case a: AppliedType => a.args
       case _              => List.empty[TypeRepr]
 
-    // Build a table for resolving type parameters, e.g., class MyClass[A, B]  -> Map("A" -> TypeRepr, "B" -> TypeRepr)
-    (classTypeParams zip classTypeArgs).map { (paramType, argType) =>
-      paramType.name -> argType
-    }.toMap[String, TypeRepr]
+    (classTypeParams zip classTypeArgs)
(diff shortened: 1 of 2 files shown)


OndrejSpanel:
Note: like #3466 this avoid using `.tree` and provides a simpler implementation of `typeMappingTable`. If you do not want to implement the Scala 2 solution of the issue, I suggest disabling the test and merging the fix anyway, or merging at least the `typeMappingTable` change.

xerial:
As we don't need to support everything for Scala 2 as well, we can put the test case under src/test/scala-3 folder to pass the CI

OndrejSpanel:
Rebased, test moved to Scala 3. Should be ready to go.

Issue #3417, opened 2024-02-27
Surface.of sometimes fails for classes using generic types on parameters


OndrejSpanel:
For information: I have reported a Scala 3 issue which sometimes causes `Surface` to fail with an error on some classes using generic types for parameters.

See https://github.com/lampepfl/dotty/issues/19795

The failure can be seen even on simple classes, like:

```scala
  trait MyOption[T]

  class Wrap(val option: MyOption[Int])
```

PR #3466, opened 2024-03-22
surface (internal): Use .typeMember instead of .tree to get parameter types

OndrejSpanel:
Do not use `.tree` to obtain parameter types, use `typeMember` instead. The code is simpler and more reliable, avoiding issue https://github.com/scala/scala3/issues/19795 (#3417)


xerial:
Thanks for the improvement. Merged

PR #3482, opened 2024-04-11
sql (fix): Digit identifier must be quoted

takezoe:


airframe-sql/src/main/scala/wvlet/airframe/sql/model/Expression.scala
```diff
@@ -527,8 +527,8 @@ object Expression {
     override lazy val resolved: Boolean             = true
   }
   case class DigitId(value: String, nodeLocation: Option[NodeLocation]) extends Identifier {
-    override def sqlExpr: String  = value
-    override def toString: String = s"Id(${value})"
+    override def sqlExpr: String  = s""""${value}""""
+    override def toString: String = s"""Id("${value}")"""
```
xerial, line 531: No need to quote this string representation

xerial:
@takezoe I think DigitIdentifiers are intended for representing `GROUP BY 1, 2, 3 ...`.  In which case do we need to quote them?

takezoe:
> I think DigitIdentifiers are intended for representing GROUP BY 1, 2, 3 ...

Ah, I see.

> In which case do we need to quote them?

If table, etc. is like below, "123" is parsed as a digit identifier
```sql
SELECT * FROM database."123"
```
and regenerated SQL from the AST will be invalid:
```sql
SELECT * FROM database.123
```
Now I think the expected behavior is that these identifiers should be parsed as quoted identifier. I will send another PR.

takezoe:
Created https://github.com/wvlet/airframe/pull/3493

PR #3493, opened 2024-04-20
sql (fix): Quote digit QName parts to generate valid SQL

takezoe:


airframe-sql/src/main/scala/wvlet/airframe/sql/model/Expression.scala
```diff
@@ -463,9 +463,6 @@ object Expression {
       QuotedIdentifier(x.stripPrefix("\"").stripSuffix("\""), None)
     } else if (x.matches("[0-9]+")) {
       DigitId(x, None)
-    } else if (!x.matches("[0-9a-zA-Z_]*")) {
-      // Quotations are needed with special characters to generate valid SQL
-      QuotedIdentifier(x, None)
```
takezoe, line 468: Moved to `QName.sqlExpr` so that this method can focus on identifier generation.

xerial:
Ok. Merged

PR #3521, opened 2024-05-02
surface (fix): Fixed a bug in getting a surface of inner opaque type

OndrejSpanel:
The PR contains a failing test demonstrating issue. Following code fails:

```scala
  object O:
    opaque type InnerOpaque = Double

  test("Opaque types from inner object") {
    val s = Surface.of[O.InnerOpaque]
  }
```

The error is:

> java.util.NoSuchElementException: key not found: wvlet.airframe.surface.Scala3NewTypeTest.O.InnerOpaque

I will try to create a fix.


OndrejSpanel:
The opaque type aliases need the same handling for inner types which normal type aliases already had.

xerial:
Thanks for the fix. Merged

PR #3533, opened 2024-05-14
surface: Simplify extractSymbol - avoid .tree to make library working with Scala 3.4.2

OndrejSpanel:
Followup to #3521

The method `extractSymbol` used to dealias inner types for both opaque types and type aliases does not work when the library is used from Scala 3.4.2 application. The reason is infamous `.tree` method again, which is known and documented to be unreliable.

I am not sure why it was used, the code seems simpler without it, but maybe I am missing something. After the simplification all tests still pass.

### DIFF
--- a/airframe-surface/src/main/scala-3/wvlet/airframe/surface/CompileTimeSurfaceFactory.scala
+++ b/airframe-surface/src/main/scala-3/wvlet/airframe/surface/CompileTimeSurfaceFactory.scala
@@ -173,14 +173,22 @@ private[surface] class CompileTimeSurfaceFactory[Q <: Quotes](using quotes: Q):
 
   private def extractSymbol(t: TypeRepr) =
     val dealiased = t.dealias
-    if t != dealiased then surfaceOf(dealiased)
+
+    println(s"dealiased ${dealiased.show} of ${t.show}")
+    println(s"  maybeOwner.declarations ${t.typeSymbol.maybeOwner.declarations.map(_.name).mkString(",")}")
+
+    if t != dealiased then
+      println(s"dealiased as ${dealiased.show}")
+      surfaceOf(dealiased)
     else
       // t.dealias does not dealias for path dependent types, so try to find a matching inner type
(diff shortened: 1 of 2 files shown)


OndrejSpanel:
> After the simplification all tests still pass.

The tests passed locally, but they do not on CI. I will check what is going on.

OndrejSpanel:
I am afraid I am unable to fix this. I was missing one dealias, once I have added it, types both with and without the fix seem correct, without the fix the type is represented as `_ >: MyBackend.this.DatabaseDef <: MyBackend.this.DatabaseDef`, which somehow prevents something which happens latter with `MyBackend.this.DatabaseDef`, the type used with the fix.

It works fine in my "light" fork without the method accessors.

Unless this is fixed, expect trouble once users of the library port to 3.4.2.

No fix logs:

```
dealiased wvlet.airframe.surface.PathDependentType.MyProfile#Backend.Database of wvlet.airframe.surface.PathDependentType.MyProfile#Backend.Database
  maybeOwner.declarations <init>,Database,DatabaseDef,DatabaseDef,DatabaseDef$
Match 1 _ >: MyBackend.this.DatabaseDef <: MyBackend.this.DatabaseDef in Some(type Database)
```

With the fix:

```
dealiased wvlet.airframe.surface.PathDependentType.MyProfile#Backend.Database of wvlet.airframe.surface.PathDependentType.MyProfile#Backend.Database
  maybeOwner.declarations <init>,Database,DatabaseDef,DatabaseDef,DatabaseDef$
Match 1 type Database in Some(type Database) as MyBackend.this.Database -> MyBackend.this.DatabaseDef
```

Note `MyBackend.this.DatabaseDef`, which reminds of #3411. The error itself reminds of #3440

xerial:
As a preliminary work, I've added a CI for Scala 3.4.2 to reproduce this issue #3590. As path-dependent type support is a bit complicated, I will revisit the current implementation later based on this PR. Thanks for reporting and trying to fix the issue.

xerial:
#3669 adopted this approach. Instead of handling the complicated ownership issue, I've fixed the test case side so that we don't need to address the outer reference for path-dependent types, which will not be used so frequently.

PR #3580, opened 2024-07-07
airframe-ulid: Add toUUID/fromUUID to ULID

victornguen:
Add functions for converting ULID to Java UUID and vice versa.


xerial:
@victornguen Thanks! I've checked the conversion behavior is the same with https://www.ulidtools.com/

PR #3623, opened 2024-09-02
Fix macros under Scala 3.6 change to context bounds

WojciechMazur:
Since Scala 3.6  context bounds would be mapped to using arguments, previously they were using implicit arguments. 
This change affects the macros which eliminate ClassTag context bound arguments. 

The issue was detected by the Scala 3 Open Community build - [build logs](https://github.com/VirtusLab/community-build3/actions/runs/10640209328)


xerial:
@WojciechMazur Thank you for fixing that! I appreciate the effort put into maintaining the community build, which makes proactive fixes like this possible.

PR #3648, opened 2024-09-24
Upgrade logback-core to 1.5.8

xerial:
SizeAndTimeBasedFNATP of logback-core was deprecated 8 years ago.  This fix will remove its usage https://github.com/qos-ch/logback/blame/42caff87ae6eb553dcbf77e25ba87a9d340357ee/logback-core/src/main/java/ch/qos/logback/core/CoreConstants.java#L240


xerial:
set-integration test fails because the old log back-core dependency is somehow still used. I'll release a new version first to pass the test

armanbilge:
FYI this change dropped support for Java 8.

PR #3744, opened 2024-12-02
sql (fix): Support POSITION

takezoe:


xerial:
@takezoe Thanks. I'll merge this fix and release a new version. 

Soon, I'd like to use revised parser/logial plan node classes of Wvlet, which supports Scala.js/Native as well: 
https://github.com/wvlet/wvlet/tree/main/wvlet-lang/src/main/scala/wvlet/lang/compiler/parser

The current airframe-sql code needs to be ported to our private project. I can help that if you are ok.

takezoe:
Oh, I see. Let's talk internally.

PR #3921, opened 2025-05-26
sql (fix): Support special date and time functions

takezoe:


xerial (approved):

airframe-sql/src/main/scala/wvlet/airframe/sql/parser/SQLGenerator.scala
```diff
@@ -463,6 +463,8 @@ object SQLGenerator extends LogSupport {
         s"(${values.map(printExpression).mkString(", ")})"
       case Parameter(index, _) =>
         "?"
+      case e: Expression =>
+        e.sqlExpr
```
takezoe, line 467: It might be better to cover concrete expression types one by one rather than covering all types here to notice unintentional SQL generation?

PR #3929, opened 2025-05-30
sql (fix): Fix SQL generation for POSITION IN subquery

takezoe:


airframe-sql/src/test/scala/wvlet/airframe/sql/parser/SQLGeneratorTest.scala
```diff
@@ -203,6 +203,14 @@ class SQLGeneratorTest extends AirSpec {
     sql shouldBe "SELECT xid FROM A WHERE POSITION('str' IN xid) > 0"
   }
 
+  test("print POSITION IN (subquery)") {
+    val resolvedPlan =
+      SQLAnalyzer.analyze("SELECT xid FROM A WHERE POSITION('str' IN (SELECT xid FROM B)) > 0", "default", demoCatalog)
```
takezoe, line 208: Generated query was like this without this fix:
```
SELECT xid FROM A WHERE POSITION('str' IN SubQueryExpression(Project[xid:? := Id(xid)](TableScan(name:B, table:default.B, columns:[])),Some(NodeLocation(1,43)))) > 0
```

airframe-sql/src/test/scala/wvlet/airframe/sql/parser/SQLGeneratorTest.scala
```diff
@@ -203,6 +203,14 @@ class SQLGeneratorTest extends AirSpec {
     sql shouldBe "SELECT xid FROM A WHERE POSITION('str' IN xid) > 0"
   }
 
+  test("print POSITION IN (subquery)") {
+    val resolvedPlan =
+      SQLAnalyzer.analyze("SELECT xid FROM A WHERE POSITION('str' IN (SELECT max(xid) FROM B)) > 0", "default", demoCatalog)
```
takezoe, line 208: Generated query was like this without this fix:
```
SELECT xid FROM A WHERE POSITION('str' IN SubQueryExpression(Project[?:? := FunctionCall(max, Id(xid), distinct:false, window:None)](TableScan(name:B, table:default.B, columns:[])),Some(NodeLocation(1,43)))) > 0
```
Because `Position.sqlExpr` is used for generating SQL:

https://github.com/wvlet/airframe/blob/ab3f07e514a1c8efb16b4190a4b41c07360072a3/airframe-sql/src/main/scala/wvlet/airframe/sql/model/Expression.scala#L1349-L1350

PR #4012, opened 2025-08-15
Mention Gradle multi-backend Scala plugin.

dubinsky:
**Description**

Mention support for AirSpec in the Gradle multi-backend Scala plugin.

**Checklist**

- [x] This pull request focuses on a single task.
- [x] The change does not contain security credentials


dubinsky:
> Thanks for the update!

Thank you!!

xerial:
That's no problem. This error is shown because the doc has a Docusaurus specific header

